PR #3, opened 2018-03-16
Autogen handlers

Calebjh:
Object level C/R/U/D barebones GORM. No service/RPC direct integration just yet.

No end-to-end test just yet, but basic associations may work already, so long as they comply to GORM conventions (foreign key named as foreignTypeID), or are explicitly gorm tagged.


README.md
```diff
@@ -31,33 +33,50 @@ make
...
 
-You can run `make install` or `go install` directly.
+##### Installation
+
+To use this tool, install it from code with `make install`, `go install` directly,
+or `go get github.com/infobloxopen/protoc-gen-gorm`.
+
+### Usage
+
+Once installed, the `--gorm-out=.` or `--gorm-out=${GOPATH}src`
```
johnbelamaric: s/gorm-out/gorm_out/

README.md
```diff
@@ -31,33 +33,50 @@ make
...
+option can be specified in a protoc command to generate the .pb.gorm.go files.
+
+Any types with the `option (orm.opts).ormable = true` will have the following
```
johnbelamaric: Change `orm` to `gorm`, since this is really a gORM-specific plugin.

example/contacts/contacts.proto
```diff
@@ -0,0 +1,80 @@
...
+message GetRequest {
+    uint64 id = 1;
+}
+
+message SMSRequest {
+    uint64 id = 1;
+    string message = 2;
+}
+
+service Contacts {
```
johnbelamaric: This is ok for now, I think though that we want to nail down the conventions. @dmacthedestroyer suggested CreateRequest, for example. So, ultimately this example should follow those conventions (or we should just have the one example app and not put all the examples in here at all).

Something like:

```
message Contact {
...
}

message CreateRequest {
  Contact contact = 1;
  // maybe some other control or metadata, for example, a certain app may want to add
  // options to the create
  bool overwrite = 2; // overwrite an existing contact with the same email address
  // or add tags
  TagSet tags = 3;
}

message CreateResponse {
  Contact contact = 1;
}

service Contacts {
...
```

We need to have a separate discussion about what we want those basic calls to look like. They should be built in a way that they are extensible and integrate with the other Atlas services.

glide.lock
```diff
@@ -1,5 +1,5 @@
 hash: 5d4fb27f861fb4a53f3cfafc8916cba71c7038b3438b57c24ffa36f028689cb2
-updated: 2018-03-06T17:28:16.140271-08:00
+updated: 2018-03-14T14:32:50.982036-07:00
```
johnbelamaric: Please move to dep in a separate PR (I created an issue for this).

example/contacts/contacts.pb.gorm.go
```diff
@@ -0,0 +1,98 @@
...
+type ContactORM struct {
+	ID           uint64
+	FirstName    string
+	MiddleName   string
+	LastName     string
+	EmailAddress string
+}
+
+func (ContactORM) TableName() string {
+	return "Contact"
```
johnbelamaric: I think the gORM conventions would make this `contacts`, wouldn't it?

README.md
Calebjh: Should ormable -> gormable, too?

johnbelamaric: if the overall "orm" is changed to "gorm" then no, that's not needed.

glide.lock
Calebjh: Will do

PR #8, opened 2018-03-23
add multitenant support

dmacthedestroyer:
If the `multi_tenant` flag is set in the `.proto`, a `ID` field will be
added to the gorm model, and the CRUD methods will be modified to
include tenancy checks.

*Limitations:* if an `id` field isn't present on the message, then the
`DefaultUpdateX` function can't be created. This could be extended to
have better support, but for this iteration the `id` convention will be
required.


plugin.go
```diff
@@ -376,15 +379,26 @@ func (p *ormPlugin) generateDefaultHandlers(file *generator.FileDescriptor) {
...
+		p.generateCreateHandler(message)
+		p.generateReadHandler(message)
+		p.generateUpdateHandler(message)
+		p.generateDeleteHandler(message)
 		// p.generateListHandler(file, service, method)
 	}
 }
 
-func (p *ormPlugin) generateCreateHandler(file *generator.FileDescriptor, message *generator.Descriptor) {
+func (p *ormPlugin) GetOption(message *generator.Descriptor) *gorm.GormMessageOptions {
```
Calebjh: Message options are the most common, but is "GetOption" descriptive enough, given the number of possible option types?

dmacthedestroyer: that's a good call.. I keep forgetting Go doesn't have method overloading

PR #12, opened 2018-03-29
UUID support

Calebjh:
Adding support for UUIDValue type (included in types/types.proto) and google.protobuf.Timestamp WKT. 

To accommodate this, also adding error returns on converter functions.

Also fixed an issue where the slice was repeatedly checked against nil instead of the elements of the slice being checked as they were handled.


plugin.go
```diff
@@ -252,7 +287,7 @@ func (p *ormPlugin) generateMapFunctions(message *generator.Descriptor) {
 		}
 		p.generateFieldMap(message, field, false)
 	}
-	p.P(`return to`)
+	p.P(`return to, err`)
 	p.P(`}`)
```
amaskalenka: If the last field of a message `is not an ORMable message type` we have a compilation error
```console
 58:6: expected '(', found 'IDENT' ConvertACLFromORM (and 2 more errors)
--gorm_out: protoc-gen-gorm: Plugin failed with status code 1.
```

That is caused by missing `}` symbol for the `for` statement.

Calebjh: Good catch. Specifically for repeated nonORMable fields. I'll fix that.

Calebjh: Also missed the repeated UUID and Timestamp cases. Fixing those, too.

PR #13, opened 2018-03-29
First attempt to implement default server generation

dkukharau:
Hi all,

Summary of the changes:

- Remove unnecessary staff from `main.go` 
- Cleanup Makefile
- Add default server generation

Default server generation notes:
- In order to enable default server generation you need to set Protobuf Service level option, e.g.:
```protobuf
service Point {
  rpc CreateIntPoint ( int_point ) returns ( int_point ) {}
  rpc ReadIntPoint ( int_point ) returns ( int_point ) {}
  rpc UpdateIntPoint ( int_point ) returns ( int_point ) {}
  rpc ListIntPoint ( google.protobuf.Empty ) returns ( int_point_page ) {}
  rpc DeleteIntPoint ( int_point ) returns  ( google.protobuf.Empty ) {}
}
```
- Only CRUD methods are generated for now. Probably we'll need to add generation of stubs for other methods as well(e.g. SendSMS) if we want to end up with out-of-the-box working GRPC server(We'll not be able to register a service if some methods of the service interface are not implemented).
- Generated List method relies on the condition that there is `Results` field in returned message type(probably we can do it more flexible, but I did it this way for simplicity)
- Proto Comments are not handled.

A few words on current protoc-gen-gorm state:
- The code readability is poor. We need to pay more attention to the code quality if we want to involve several developers to work on this tool. Though there is not much of functionality yet, it was extremely difficult to get into and make something useful here. It became even more difficult because of `protoc-gen-gogo`, which our generator relies on, as it's api has very poor documentation. Probably we could create some guides about it for ourselves in future.
- Why three examples(proto files)? I think we need to keep one and fulfill it with all the features supported by generator. Surely, we should end up with compilable and working example.


plugin.go
```diff
@@ -559,3 +562,93 @@ func (p *ormPlugin) generateListHandler(message *generator.Descriptor) {
...
+					methodName := lintName(generator.CamelCase(method.GetName()))
+					if strings.HasPrefix(methodName, "Create") {
+						p.generateCreateServerMethod(file, service, method)
+					} else if strings.HasPrefix(methodName, "Read") {
+						p.generateReadServerMethod(file, service, method)
+					} else if strings.HasPrefix(methodName, "Update") {
+						p.generateUpdateServerMethod(file, service, method)
+					} else if strings.HasPrefix(methodName, "Delete") {
+						p.generateDeleteServerMethod(file, service, method)
+					} else if strings.HasPrefix(methodName, "List") {
```
dmacthedestroyer: if possible, it would be nice to check if the `.proto` service return message actually has a `Results` field of the proper type and render something that compiles either way. This current approach means not fitting the convention results in code that doesn't compile. This would be nice to do for all of these methods.

dkukharau: I agree and it seems that it's possible to implement that, though it's not trivial. Along with return type we need to validate method parameter as well, so I would like to come back to this when the positive cases are 100% working. This server is generated only if `(gorm.server).autogen` option is explicitly set, so the chance that the user knows what he is doing increases. 
In case of failed validation we may just leave `return nil, nil` or `return nil, &NotImplementedError{"MethodName"} as a default implementation, that will always compile.

dkukharau:
- Added generation of stubs for custom methods
- Added response and request wrappers. Implementation relies on the existence of `result`, `results` and `payload` fields in corresponding wrappers.  Checkout test2.proto.

dkukharau:
Hey @dmacthedestroyer, could you merge the PR? Don't have merge privileges here.

PR #16, opened 2018-04-06
Bug fixes

Calebjh:
Fixes some issues observed by Andrei for type names being improperly cased and for nested objects having underscores in the ORM typenames that shouldn't have been there.

Switches from panicking interface type assertions to get message options to erring assertions, hadn't noticed any problems from this, but it should be safer.

UUIDValue special type now become a *uuid.UUID instead of non-pointer to match better with the WKTs and prevent an automatic nil check against a non-pointer type if used as a PK/FK.

Now also checks in ORM-only fields for association keys.


plugin.go
```diff
@@ -810,14 +821,35 @@ func (p *ormPlugin) removeChildAssociations(message *generator.Descriptor) bool
...
 			for _, childField := range typeNames[rawFieldType].GetField() {
 				if strings.EqualFold(childField.GetName(), k) {
 					childFKeyTypeName, _ = p.GoType(message, childField)
 					break
 				}
 			}
+			if typeNames[rawFieldType].Options != nil {
+				ext, err := proto.GetExtension(typeNames[rawFieldType].Options, gorm.E_Opts)
+				opts, valid := ext.(*gorm.GormMessageOptions)
+				if err == nil && valid {
```
dmacthedestroyer: I'm not sure what `err` means in this context, but I'm guessing it's not that important?

PR #20, opened 2018-04-09
update dependencies, add test Makefile target

dmacthedestroyer:


Makefile
```diff
@@ -22,3 +22,10 @@ example: default
 		-I$(SRCPATH)/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
 		--go_out="plugins=grpc:$(SRCPATH)" --gorm_out="$(SRCPATH)" \
 		example/contacts/contacts.proto
+
+	goimports -w ./example
```
Calebjh: I would prefer to have the goimports called in the test target instead of the example target, so that it remains clear what exactly is being generated by the tool.

PR #26, opened 2018-04-12
Improve default server generation

dkukharau:
Any services with the `option (orm.server).autogen = true` will have basic grpc server generated:

- For service methods with names starting with `Create|Read|Update|Delete` generated implementation will call basic CRUD handlers.
- For other methods `return &MethodResponse{}, nil` stub is generated.

For CRUD methods to be generated correctly you need to follow specific conventions:
- Ormable Type should be wrapped in Request/Response messages under `payload`, `result` or `results` fields according to the method type. Check out [example](example/feature_demo/test2.proto).
- Read and Delete methods should have an `id` field in Request message.
- For Delete method pass `(gorm.method).object_type` option to indicate which Ormable Type it's related to. Check out [example](example/feature_demo/test2.proto).

If conventions are not met stubs are generated for CRUD methods.


dkukharau:
Removed goimports from `test` target as it's better for us to generate code which works without the help of `goimports`. Examples are working fine.

Evgeniy-L:
Daniil, please resolve conflicts

plugin/servergen.go
```diff
@@ -45,61 +44,217 @@ func (p *OrmPlugin) generateDefaultServer(file *generator.FileDescriptor) {
...
 	inType, outType, methodName, svcName := p.getMethodProps(service, method)
 	p.generateMethodSignature(inType, outType, methodName, svcName)
-	p.P(`var out `, p.TypeName(outType))
-	p.P(`res, err := DefaultRead`, strings.TrimPrefix(methodName, "Read"), `(ctx, in.GetPayload(), m.DB)`)
-	p.P(`out.Result = res`)
-	p.P(`return &out, err`)
-	p.P(`}`)
+	follows, typeName := p.followsReadConventions(inType, outType)
+	if follows {
+		p.P(`res, err := DefaultRead`, typeName, `(ctx, &`, typeName, `{Id: in.GetId()}, m.DB)`)
```
Calebjh: The choice to use golint compliant names for the ORMable objects is causing way more trouble than I anticipated.
In these statements, the first typeName is linted (since I thought of the methods as CURDLing an ORM object) and the second one is not linted, since it's coming from the pb generated file. This will only make a difference on names with initialisms though (ACL, ID, API, etc.) so it's a horrible 'gotcha'.

plugin/servergen.go
```diff
@@ -45,61 +44,217 @@ func (p *OrmPlugin) generateDefaultServer(file *generator.FileDescriptor) {
...
+	inMsg := inType.(*generator.Descriptor)
+	var hasID bool
+	for _, field := range inMsg.Field {
+		if field.GetName() == "id" {
+			hasID = true
+		}
+	}
+	var typeName string
+	v, err := proto.GetExtension(method.GetOptions(), gorm.E_Method)
+	opts := v.(*gorm.MethodOptions)
```
Calebjh: This line panics if (gorm.server).autogen is set, but (gorm.method).object_type is not set. Use the opts, ok := v.(*gorm.MethodOptions) to avoid this. If it needs to fail in that scenario, something like p.Fail("Cannot generate Delete handler if object_type option is not set in rpc definition") would be good.

PR #27, opened 2018-04-13
add full go package name to proto file

dmacthedestroyer:


example/feature_demo/test.pb.gorm.go
```diff
@@ -41,7 +41,7 @@ import uuid "github.com/satori/go.uuid"
 import fmt "fmt"
 import math "math"
 import google_protobuf1 "github.com/golang/protobuf/ptypes/wrappers"
-import google_protobuf2 "github.com/golang/protobuf/ptypes/empty"
+
```
dmacthedestroyer: this change occurred when I ran the `make example` task with no changes -- not sure what's different, but I'm assuming it was just out of date.

dmacthedestroyer: oops, nevermind.. it's a result of the `goimports` command being on the `make test` target. I'm going to remove that command from the `test` target so this doesn't happen.

PR #29, opened 2018-04-14
Improve default server generation

dkukharau:
Accidentally merged #26  by pushing to upstream, reverted the changes, recreate the PR:

Addressed comments of @Calebjh, thanks, good catches!


dkukharau:
@Calebjh @dmacthedestroyer Regarding name linting.. I don't see much value from them in this context, but it really complicates the things, especially it could be a big trouble for somebody unfamiliar with the tool coming here and trying to do something.. Maybe we can consider dropping this approach?

johnbelamaric:
I would agree that name linting in this context isn't worth the hassle it seems to be creating.

Makefile
```diff
@@ -14,7 +14,7 @@ install:
 example: default
 	protoc -I. -I$(SRCPATH) -I./vendor \
 		--go_out="plugins=grpc:$(SRCPATH)" --gorm_out="$(SRCPATH)" \
-		example/feature_demo/*.proto
+		example/feature_demo/test.proto example/feature_demo/test2.proto
```
Calebjh: Wondering because this came up in a discussion earlier, did you have errors with the *.proto notation, or is this just clearer?

dkukharau: Not an error, but a little inconsistency across environments. This `example/feature_demo/*.proto` may expand either into `example/feature_demo/test.proto example/feature_demo/test2.proto` or `example/feature_demo/test2.proto example/feature_demo/test.proto` depending on the environment, and thus results in a slightly different generated code(mainly comment headers differ).

Issue #28, opened 2018-04-13
1:many convention breaks with snake_case parent key name


dmacthedestroyer:
the following proto message:
```
message Contact {
    option (gorm.opts) = {ormable: true};
    uint64 id = 1;
    repeated Address addresses = 6;
}

message Address {
  option (gorm.opts).ormable = true;
  uint64 id = 1;
  uint64 contact_id = 2;
}
```
fails with the following error:
```
protoc-gen-gogo: error:Child type Address seems to have no foreign key field linking it to parent type Contact : expected field ContactID in Address associated with field ID in Contact
--gorm_out: protoc-gen-gorm: Plugin failed with status code 1.
make: *** [protobuf] Error 1
```

changing the `  uint64 contact_id = 2;` to be `contactid` (no underscore) gets it to compile properly, but this doesn't comply to the naming best practices for proto files

Calebjh:
Should be resolved in extensive changes from #31

PR #31, opened 2018-04-17
Forsake lint

Calebjh:
Remove all type and field name Linting. 
Resultant changes:
- pbTypeNames and ormTypeNames (less ORM suffix) are now equal, makes it simpler to unify them and remove the unnecessary utility function getTypeNames
- to and from field names in conversions are now the same (was fieldName and fromName respectively)

Also made example file type names convention compliant to test how ID fields function and increased use of getMessageOptions helper function to remove wasted code


plugin/handlergen.go
```diff
@@ -294,22 +290,22 @@ func (p *OrmPlugin) removeChildAssociations(message *generator.Descriptor) bool
...
 		keys := p.findAssociationKeys(message, typeNames[typeName], field)
 		childFKeyTypeName := ""
-		_, _, fieldTypeNameORM := getTypeNames(typeNames[rawFieldType])
-		p.P(`filterObj`, rawFieldType, ` := `, fieldTypeNameORM, `{}`)
+		fieldTypeName := p.TypeName(typeNames[rawFieldType])
+		p.P(`filterObj`, rawFieldType, ` := `, fieldTypeName, `ORM{}`)
 		for k, v := range keys {
 			for _, childField := range typeNames[rawFieldType].GetField() {
-				if strings.EqualFold(childField.GetName(), k) {
+				if strings.EqualFold(generator.CamelCase(childField.GetName()), k) {
```
dmacthedestroyer: what's the purpose of `CamelCase` if you're comparing ignoring case? I'm guessing it translates snake case?

Calebjh: Exactly. This should actually be the fix for issue #28 for cases where the ID fields are defined in snake case, camel case, camel case with initialisms, or even StUdLyCaPs.

PR #33, opened 2018-04-17
update to use version 0.2.0 of infobloxopen/atlas-app-toolkit

dmacthedestroyer:
version 0.2.0 renamed the `auth/GetTenantID` function to `GetAccountID`


dmacthedestroyer:
@Calebjh I was wondering about that... do you think the message option should be renamed to `multi-account`? That seems reasonable to me.

Calebjh:
Hmm. I think it should be changed to match (multi_account). Obviously multi-accountancy sounds wrong, but it would also be weird to not have the option match.

Evgeniy-L:
I think that we should try to be consistent in names to avoid any type of confusion. I agree with @Calebjh that "tenant" should be replaced with "account".

PR #38, opened 2018-04-30
Add hooks by interface, change converters to receiver functions

Calebjh:
Add the hook architecture for before/after conversions via interfaces. (Naming choices are up for discussion)

Also switched the form of the converter functions to ToPB and ToOrm on the object types. However, we can only use this for interfaces if we force the return type to some interface like `(interface{}, error)` instead of the actual types (unless I'm missing something?), making it more complicated to use. Do we have an actual use case where these converter functions being interfaces would be useful, and if they aren't interface functions is it worth switching the form here?


types/hooks.go
```diff
@@ -0,0 +1,25 @@
+package gormable_types
+
+// WithBeforeToOrm has a hook function to be called before an Ormable type
+// is converted ToOrm
+type WithBeforeToOrm interface {
+	BeforeToOrmHook(interface{})
```
dmacthedestroyer: hmm.. the empty interface is kind of a bummer. Could we at least make it `pb.Message` for the ones that take in a proto message? I don't think we have any kind of interface for the ones that take an ORM object.

johnbelamaric: Maybe I am missing something, I'd like to see exactly how these are used. I am thinking that since we're generating the code anyway. Why wouldn't we just generate interfaces with the proper signatures? The developer then would create the methods with the right receiver and be done with it.

An alternative, if we want to allow multiple hooks per type, would be to track them as callbacks. However, instead of just implementing the methods then, they would have to register them which is a pain.

dmacthedestroyer: pre- and post-hooks generated by each type might not be too bad, except for the really weird interface names that would have to generate. That probably isn't so bad, either, though, since as far as the implementation is concerned the function name is just `BeforeToOrmHook`. I agree -- let's try it out that way unless Caleb has some insight.

We shied away from the multiple hooks per type because of the whole having to register them thing... if someone needs that functionality, they can wire that behavior into their own hooks and register them as they see fit (in `main.go`, in an `init` func, etc). If it's a really common thing, we can build out a library for it.

As for naming, let's go with `Before/AfterToOrm/Pb`, without the `Hook` in there.

johnbelamaric: Yes, agreed. No `Hook`

PR #41, opened 2018-05-02
Add gorm example for consideration.

Calebjh:
A candidate to replace some of the other examples in this repo from the gorm example at http://doc.gorm.io/models.html.

Should be corrected/simplified as the association handling is implemented.

Also demonstrates the hook behavior for a synthetic field


example/user/user.proto
```diff
@@ -0,0 +1,73 @@
...
+message User {
+  option (gorm.opts) = {
+    ormable: true,
+    include: [
+    {type: "int32", name: "billing_address_id"},
+    {type: "int32", name: "shipping_address_id"}
+    ]
+  };
+  int32 id = 1;
+  google.protobuf.Timestamp created_at = 2;
```
dmacthedestroyer: these can generally be excluded from the proto file, correct? I'm thinking we shouldn't expose them via the API

johnbelamaric: Why not? I could be useful meta data in some cases. Certainly for something like audit log entries...

dmacthedestroyer: sorry, I should have been more specific: We can include them, but generally they won't need to be included.

I think that in this particular canonical example, we should demonstrate that if you don't specify the `created_at` and `updated_at` fields then they're still included in the database.

so I'm thinking we shouldn't expose them in this example.

Makefile
```diff
@@ -20,6 +20,10 @@ example: default
 		--go_out="plugins=grpc:$(SRCPATH)" --gorm_out="$(SRCPATH)" \
 		example/feature_demo/test.proto example/feature_demo/test2.proto
 
+	protoc -I. -I$(SRCPATH) -I./vendor \
+		--go_out="plugins=grpc:$(SRCPATH)" --gorm_out="$(SRCPATH)" \
+		example/user/user.proto
```
dmacthedestroyer: I think this should replace the `contacts` example.

example/user/user.override.go
```diff
@@ -0,0 +1,7 @@
+package user
+
+import "time"
+
+func (m *UserORM) AfterToPB(user *User) {
```
dmacthedestroyer: Add documentation to demonstrate what this does, so that if someone were to read this code they'd know it's a means to supply additional customization.

The hope here is that if someone comes to us and asks, "how do I add synthetic fields?" we can point them to this file and they'd understand.

PR #43, opened 2018-05-07
Rename gormable types package, rename example, update Makefile

Calebjh:
Change the protobuf package name from `gormable_types` to `gorm.types` and go package name to `types` to match conventions (no underscores). 

Change the file names for the example files to be more informative.

Update the Makefile to match the changes.


Makefile
```diff
@@ -10,15 +10,15 @@ build:
...
+	protoc --go_out=$(SRCPATH) types/types.proto
 
 install:
 	go install
 
 example: default
 	protoc -I. -I$(SRCPATH) -I./vendor \
 		--go_out="plugins=grpc:$(SRCPATH)" --gorm_out="$(SRCPATH)" \
-		example/feature_demo/test.proto example/feature_demo/test2.proto
+		example/feature_demo/demo_types.proto example/feature_demo/point_service.proto
```
dmacthedestroyer: `test` and `test2` are indeed pretty bad names, but I'm not sure `demo_types` and `point_service` are any better.

What is `point_service`, anyway? Can it be called `demo_service`?

Also, if we're going through the trouble of trying to make these actually readable by someone, we should add line-by-line documentation on each of the fields and messages in these files to demonstrate what they should be doing.

Calebjh: Alright. I'll expand on the comments and go with `demo_service` instead. 
I'm also making Readme updates, you want me to add these here, or keep those for a separate PR?

dmacthedestroyer: I'm not opinionated about that -- whatever makes sense to you

example/feature_demo/demo_service.proto
```diff
@@ -38,25 +49,38 @@ message UpdateIntPointResponse {
 }
 
 message DeleteIntPointRequest {
-        uint32 id = 1;
+    // Only the id is needed for a delete request
+    uint32 id = 1;
 }
 
+// By convention, on DELETE no response is expected
```
dmacthedestroyer: can we still include an empty `DeleteIntPointResponse` message? I'd prefer to set that as the pattern than an empty response

example/feature_demo/demo_service.proto
```diff
@@ -38,25 +49,38 @@ message UpdateIntPointResponse {
...
 
+// A dummy type to demo an rpc that can't be autogenerated
 message Something {
 }
 
 service IntPoint {
+  // This option tells protoc-gen-gorm to generate the calls and stubs
   option (gorm.server).autogen = true;
+  // The convention names are Create/Read/Update/List/Delete. Since no type is
+  // specified, only one set of CURDL operations can be defined in each service
```
dmacthedestroyer: is this true? the type is inferred from the request/response wrapper, so it seems like you could have a `CreateFoo` and `CreateBar` and both could get generated.

example/feature_demo/demo_types.proto
```diff
@@ -9,38 +9,71 @@ import "google/protobuf/timestamp.proto";
...
-// test_types is a message that serves as an example
+// test_types is a message that includes a representative sample of the
+// available types
 message TestTypes {
+  // The 'ormable' option flags this for converter autogeneration
+  // The table option allows for overriding the default table name 'test_types'
   option (gorm.opts) = {ormable:true, table: "smorgasbord"};
+  // the (gorm.field).drop option allows for setting a field to be API only
   string api_only_string = 1 [(gorm.field).drop = true];
+  // repeated raw types are currently unsupported
```
dmacthedestroyer: ... so this field will be skipped in the GORM model

example/feature_demo/demo_types.proto
```diff
@@ -50,6 +83,9 @@ message MultiaccountTypeWithID {
 }
 
 message MultiaccountTypeWithoutID {
+  // here no id field is given, but it is still a multitenant type
+  // this is a bad pattern, as there is no way to uniquely identify objects
+  // in the DB, so R/U/D behavior is not well defined
```
dmacthedestroyer: will this cause generation, or result in no GORM model for the type?

example/feature_demo/demo_types.proto
```diff
@@ -58,10 +94,16 @@ message MultiaccountTypeWithoutID {
...
 message APIOnlyType {
+  // here the ormable flag is not used, so nothing will be generated for this
+  // object at the ORM level
   string contents = 1;
 }
 
 message TypeBecomesEmpty {
+  // here the object is ormable, but the only contents are a non-ormable object.
+  // this will result in an empty object at the ORM level. This is not a
+  // suggested use case, but could be used if you wanted a fully synthetic
```
dmacthedestroyer: hmm.. should we be publishing unsuggested use cases?

example/feature_demo/demo_types.proto
Calebjh: Currently it will generate, and but will cause issues compile time once associations get involved. Should we require that there is a field named 'id' or with a gorm tag of 'primary_key'? If yes, should that be a plugin failure condition, or just a 'drop this object' condition?

PR #46, opened 2018-05-10
fix broken imports from toolkit refactor

dmacthedestroyer:
This should be merged and have a new release once the changes to the toolkit get released


example/feature_demo/demo_types.pb.gorm.go
```diff
@@ -254,10 +254,9 @@ type TypeWithIDWithAfterToPB interface {
 
 // MultiaccountTypeWithIDORM demonstrates the generated multi-account support
 type MultiaccountTypeWithIDORM struct {
-	AccountID                   string
-	Id                          uint64
-	SomeField                   string
-	MultiaccountTypeWithoutIDId uint32
```
dmacthedestroyer: I'm assuming this field was removed in a previous commit but the file wasn't synced prior to merging?

dmacthedestroyer:
I was thinking of running `dep ensure` but we currently don't even reference the toolkit in the lock file. Should we?

PR #48, opened 2018-05-11
Fix a regression error where tablename no longer used gorm option

Calebjh:
Oops


dmacthedestroyer:
Oops, indeed... time for automated tests!

PR #50, opened 2018-05-17
Has-One/Has-Many/Ordered-Has-Many/GORM tags

dkukharau:
Hi all,
Here is a bunch of features I have been working recently.
I'm sorry for doing a big change in a single PR! 

Summary:
The following items are implemented according to [link](https://docs.google.com/document/d/1Y6WwgPbwuQ6AHHSHcN7mRMtydYE20c0GxLC6m4BxT6c/edit?usp=sharing).

- Has-One
- Has-Many
- Ordered-Has-Many
- More user-friendly way to customize GORM tag on generated models
- A portion of light refactoring

Will add another one example to specifically demo associations and different combinations of options. For now check out `user` example.


options/gorm.proto
```diff
@@ -40,8 +40,41 @@ extend google.protobuf.FieldOptions {
...
+    optional string type = 2;
+    optional int32 size = 3;
+    optional int32 precision = 4;
+    optional bool primary_key = 5;
+    optional bool unique = 6;
+    optional string default = 7;
+    optional bool not_null = 8;
+    optional bool auto_increment = 9;
+    optional string index = 10;
+    optional string unique_index = 11;
```
Calebjh: Some of these options are used like DDL (index, auto_increment, default), since we're advocating manually written DDL should we keep these fields or not even include them?

plugin/associations.go
```diff
@@ -0,0 +1,133 @@
...
+		assocKey, ok = parent.Fields[assocKeyName]
+		if !ok {
+			p.Fail("Missing", assocKeyName, "field in", parent.Name, ".")
+		}
+	}
+	hasMany.AssociationForeignkey = &assocKeyName
+	foreignKey := &Field{Type: assocKey.Type, GormFieldOptions: &gorm.GormFieldOptions{Tag: hasMany.GetForeignkeyTag()}}
+	var foreignKeyName string
+	if foreignKeyName = hasMany.GetForeignkey(); foreignKeyName == "" {
+		foreignKeyName = fmt.Sprintf(fieldName + typeName + assocKeyName)
```
Calebjh: This longer generated name should prevent any FK collisions, but is there an advantage to using it over the simpler gorm standard of {Foreign Type}ID? Especially since this has to match the DB Schema

plugin/associations.go
```diff
@@ -0,0 +1,133 @@
...
+	if exField, ok := child.Fields[foreignKeyName]; !ok {
+		child.Fields[foreignKeyName] = foreignKey
+	} else {
+		if exField.Type != foreignKey.Type {
+			p.Fail("Cannot include", foreignKeyName, "field into", child.Name, "as it already exists there with a different type.")
+		}
+	}
+}
+
+func (p *OrmPlugin) findPrimaryKey(ormable *OrmableType) (string, *Field) {
```
Calebjh: GORM does support compound primary keys. Is this something we want to consider here, or should we leave it at single field keys for now or indefinitely?

example/feature_demo/demo_service.pb.gorm.go
```diff
@@ -17,8 +17,6 @@ import google_protobuf2 "github.com/golang/protobuf/ptypes/empty"
 var _ = fmt.Errorf
 var _ = math.Inf
 
-// IntPointORM is a basic message type representing a single cartesian point
-// that we want to store in a database
```
Calebjh: I notice the comment is no longer propagated from the proto file to the gorm output. Is this something that is useful enough to include, or is dropping it fine? Thoughts @dmacthedestroyer

options/gorm.proto
dkukharau: If somebody's database schema is simple enough in order to build it with GORM, why we should forbid this?

Calebjh: That's fair. My concern would be that gorm has limited migration support, so if at some point the schema has to change beyond what gorm can handle like this, it would be an awkward half-gorm/half-manually written schema. I suppose that would be 'their problem' though.

dkukharau: Yeah, sure. Indeed, GORM's migration support is safe: it doesn't remove disappeared entities(columns, tables), but only creates new ones. Maybe that approach is fine if you have small databases and don't change your schema frequently. Also this can be considered as an advantage e.g. if you want to keep several versions of application working on the same database.

plugin/associations.go
dkukharau: I made it look like that at the very last moment when I heard about multi-dimensional associations. This case is listed as a [requirement](https://docs.google.com/document/d/1XLheB0qktFQbCVKANQN9MtnrCYutUYgW2_vA19YkCZ4/edit#heading=h.32x537vq59t3) from one application team. `fieldName` prefix is used to create separate foreign keys for each dimension.
Also we already have example for this in `demo_types`: 
```golang
type TypeWithIDORM struct {
	Id            uint32
	Ip            string          `gorm:"column:ip_addr"`
	SecretInt     int32           `gorm:"-"`
	Things        []*TestTypesORM `gorm:"foreignkey:ThingsTypeWithIDId;association_foreignkey:Id"`
	ANestedObject *TestTypesORM   `gorm:"foreignkey:ANestedObjectTypeWithIDId;association_foreignkey:Id"`
}
```
Now I realize that we can do it a bit more clever: say, if we have one-dimensional  association: use `typeName + assocKeyName`, if multi-dimensional: `fieldName + typeName + assocKeyName`.
Either way these are only default values and can easily be overridden.

plugin/associations.go
dkukharau: Well, I also have it in my mind. It seems for me that GORM doesn't support association through more than one field out-of-the-box, i.e. you need to use inline preload options or write custom preloading like descibed [here](http://gorm.io/docs/preload.html#Preload). 
We can actually handle this in generator, but it's a question if this will be used a lot. I would better think about how we can expose the query to a user so one can customize it in a unified way without overriding the entire handler. I know GORM hooks can do a lot for this, but I'm still not confident what actually can be customized, since GORM API is not very well documented. Another way is to write our own hooks like `After/Before-ORM/PB` and expose `*gorm.DB` there. 
Anyway, I think we should thoroughly revisit our customization approach in general since no discussion took place on this and there are a lot of similar delicate questions which need to be addressed.

example/feature_demo/demo_service.pb.gorm.go
dkukharau: Yeah, I accidentally threw away them while refactoring the code. 
I think anyway we miss comments propagation for services and methods, probably something else, fields for example.
So we can create a separate story/task for this, probably this can be handled in a unified way across different proto entities.

plugin/associations.go
dkukharau: Regarding generated files are different.. I also see this now. The reason is that golang maps don't preserve the order of keys and fields are generated in arbitrary order every time..:) Will fix that.

plugin/associations.go
dkukharau: Maybe @johnbelamaric or @dmacthedestroyer have something one this.

johnbelamaric: We don't have a need for composite keys right now. We can put it in the backlog.

example/feature_demo/demo_service.pb.gorm.go
Calebjh: I'm fine with that. Those comments are available in the proto file, having them in the gorm file as well is not so critical and can be done at a later time.

plugin/associations.go
Calebjh: That is a good point. We also have company internal use cases for that. Would a one/multi-dimensional check be a pretty quick thing to add? I'd really like for it to default to the simpler option if the simpler option works

dkukharau: Yeah, it's already in.

example/feature_demo/demo_types.pb.gorm.go
```diff
@@ -49,17 +49,15 @@ import _ "github.com/golang/protobuf/ptypes/timestamp"
...
-	OptionalString *string
-	BecomesInt     int32
-	// Empty type has no ORM equivalency
-	Uuid         *uuid.UUID `sql:"type:uuid"`
-	CreatedAt    time.Time
-	TypeWithIdId uint32
+	ANestedObjectTypeWithIDId uint32
+	BecomesInt                int32
+	CreatedAt                 time.Time
+	Numbers                   []int32
```
Calebjh: Support for primitive type slices/arrays isn't supported yet, these fields will still need to be dropped until proper handling is added.

dkukharau: Ok, fixed that.

dkukharau:
welcome @dmacthedestroyer, your turn)

plugin/plugin.go
```diff
@@ -134,90 +136,143 @@ func (p *OrmPlugin) generateMessages(message *generator.Descriptor) {
...
+			res += fmt.Sprintf("precision:%s;", string(tag.GetPrecision()))
+		}
+		if tag.GetPrimaryKey() {
+			res += "primary_key;"
+		}
+		if tag.GetUnique() {
+			res += "unique;"
+		}
+		if tag.Default != nil {
+			res += fmt.Sprintf("column:%s;", tag.GetDefault())
```
dmacthedestroyer: should be `"default"%s;"`?

dkukharau: Thanks, good catch!

PR #52, opened 2018-05-22
First pass of fixes for deleting entire table on 'zero' id value

### LINKED ISSUE #51: Using zero value for ID in delete call deletes all entries
Because of how gorm handles zero values in selections, passing the zero value into the delete handler deletes all entries in the table. This is not good. 
We should check that the input isn't the zero value.

Calebjh:
Require that the id field in Delete calls be non-zero
Don't autogenerate R/U/D handlers when there is no Primary Key for the object.
Generate stub R/U/D server functions when there is no Primary Key for the object.

Add some references to the WKT hacky override comment.

Resolves #51

### DIFF
--- a/example/feature_demo/demo_service.pb.gorm.go
+++ b/example/feature_demo/demo_service.pb.gorm.go
@@ -146,8 +146,30 @@ func DefaultDeleteIntPoint(ctx context.Context, in *IntPoint, db *gorm.DB) error
 	if err != nil {
 		return err
 	}
+	if ormObj.Id == 0 {
+		return errors.New("A non-zero ID value is required for a delete call")
+	}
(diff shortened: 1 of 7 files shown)


dkukharau:
Looks good to me.
Regarding WKT trick.. Is that needed only to hack JSON marshaler? Anything else?
In case it's the only thing: did you try to implement `JSONPBMarshaler` for `UUIDValue`? It seems to me that it's considered to be a standard way of solving our problem.

Calebjh:
Just the Marshal/Unmarshal, yes. That may be the accepted approach. I was looking for a way to "create new WKTs" instead of custom marshaling and probably missed out on a lot of other approaches because of that. I can look more into that when I have some spare cycles.

PR #53, opened 2018-05-23
Belongs-To

dkukharau:


options/gorm.proto
```diff
@@ -44,7 +44,9 @@ message GormFieldOptions {
     optional bool drop = 2;
     oneof association {
         HasOneOptions has_one = 3;
-        HasManyOptions has_many = 4;
+        BelongsToOptions belongs_to = 4;
+        HasManyOptions has_many = 5;
```
Calebjh: I believe best protobuf practice here would be to leave has_many = 4; and add the others as 5,6 instead of reassigning has_many. I don't think it matters much in this case since it's such a short time between the two versions and nobody pulled it in yet though.

dkukharau: Got it. Thanks for this note.

PR #55, opened 2018-05-24
Initial JSON type support, custom JSON type. WKT hack replacement

Calebjh:
Support for new type: gorm.types.JSONValue
Will become a glorified string wrapper at Protobuf level and convert to a postgres.JsonB type (https://github.com/jinzhu/gorm/blob/82eb9f8a5bbb5e6b929d2f0ae5b934e6a253f94e/dialects/postgres/postgres.go#L60) at the gorm level if the `driver=postgres` option is included in the `--gorm_out=` arguments, otherwise will be dropped.

Switch from WKT override-y hack thing to just implement the JSONPB Marshal/Unmarshal interfaces.


Makefile
```diff
@@ -17,7 +17,7 @@ install:
 
 example: default
 	protoc -I. -I$(SRCPATH) -I./vendor \
-		--go_out="plugins=grpc:$(SRCPATH)" --gorm_out="$(SRCPATH)" \
+		--go_out="plugins=grpc:$(SRCPATH)" --gorm_out="engine=postgres:$(SRCPATH)" \
```
dmacthedestroyer: if you don't add the `engine=postgres` part, does it default to postgres or does it crash?

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/infobloxopen/protoc-gen-gorm/types"
+)
+
+func TestUnmarshalTypes(t *testing.T) {
+	ts, _ := ptypes.TimestampProto(time.Date(2009, 11, 17, 20, 34, 58, 651387237, time.UTC))
+	marshaller := &runtime.JSONPb{OrigName: true, EmitDefaults: true}
+	for in, expected := range map[string]TestTypes{
+		`{}`: {},
+		`{"api_only_string":"important text"}`:            {ApiOnlyString: "important text"},
+		`{"numbers":[1,2,3,4]}`:                           {Numbers: []int32{1, 2, 3, 4}},
```
dmacthedestroyer: you could add more test cases here: `[]`, `nil`

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+		`{"optional_string":"something real"}`:            {OptionalString: &wrappers.StringValue{Value: "something real"}},
```
dmacthedestroyer: add test case for nil

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+		`{"becomes_int":"GOOD"}`:                          {BecomesInt: 1},
```
dmacthedestroyer: ?

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+		`{"uuid":"123abc-unchecked-uuid"}`:                {Uuid: &types.UUIDValue{Value: "123abc-unchecked-uuid"}},
+		`{"created_at":"2009-11-17T20:34:58.651387237Z"}`: {CreatedAt: ts},
```
dmacthedestroyer: bummer you had to create this `ts` variable up above. I'm happy with it the way it is, but you could also create a helper function that only returns the single value so you can put it inline here.

I've done this pattern before where things return an error but I want to put a value inline for table-driven tests like this:

```
// GetThing lives in real code, not test code, and does real things that can return real errors
func GetThing()(Thing, error){...}

// MustGetThing lives in test code, and is useful for being able to pass into things inline when you're not expecting to test the error boundary conditions. Useful for downwind tests that consume or depend on a Thing
func MustGetThing(t *testing.T) Thing {
  thing, err := GetThing()
  if err != nil {
    t.Fatal(err)
  }
  return thing
}
```

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+		`{"type_with_id_id":4}`:                           {TypeWithIdId: 4},
```
dmacthedestroyer: what happens if it gets passed a string `"4"`?

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+		`{"json_field":{"top":[{"something":1},2]}}`:      {JsonField: &types.JSONValue{Value: `{"top":[{"something":1},2]}`}},
+	} {
+		tt := &TestTypes{}
+		err := marshaller.Unmarshal([]byte(in), tt)
+		if err != nil {
+			t.Error(err.Error())
```
dmacthedestroyer: are there any error cases you want to test? If not, then you're basically never expecting to receive an error. In that case, you can use `t.Fatal(err.Error())` here instead

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+		`{}`: {},
+		`{"api_only_string":"important text"}`:            {ApiOnlyString: "important text"},
+		`{"numbers":[1,2,3,4]}`:                           {Numbers: []int32{1, 2, 3, 4}},
+		`{"optional_string":"something real"}`:            {OptionalString: &wrappers.StringValue{Value: "something real"}},
+		`{"becomes_int":"GOOD"}`:                          {BecomesInt: 1},
+		`{"uuid":"123abc-unchecked-uuid"}`:                {Uuid: &types.UUIDValue{Value: "123abc-unchecked-uuid"}},
+		`{"created_at":"2009-11-17T20:34:58.651387237Z"}`: {CreatedAt: ts},
+		`{"type_with_id_id":4}`:                           {TypeWithIdId: 4},
+		`{"json_field":{"top":[{"something":1},2]}}`:      {JsonField: &types.JSONValue{Value: `{"top":[{"something":1},2]}`}},
+	} {
```
dmacthedestroyer: fun tidbit: you can run each case in a sub-test, like this top example from the go wiki: https://github.com/golang/go/wiki/TableDrivenTests#example-of-a-table-driven-test

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+		if !reflect.DeepEqual(*tt, expected) {
+			t.Errorf("Expected unmarshaled output '%+v' did not match actual output '%+v'",
+				expected, *tt)
+		}
+	}
+}
+
+func TestMarshalTypes(t *testing.T) {
+	ts, _ := ptypes.TimestampProto(time.Date(2009, 11, 17, 20, 34, 58, 651387237, time.UTC))
+	marshaller := &runtime.JSONPb{OrigName: true, EmitDefaults: true}
```
dmacthedestroyer: do you need to consider running these tests against any other kind of marshaller? Maybe add a comment as to why you're using these settings (even if it's to say that you chose them arbitrarily)

plugin/plugin.go
```diff
@@ -17,9 +17,16 @@ const (
 	typeEnum    = 14
 
 	protoTypeTimestamp = "Timestamp" // last segment, first will be *google_protobufX
+	protoTypeJSON      = "JSONValue"
 	protoTypeUUID      = "UUIDValue"
 )
 
+// DB Engine Enum
+const (
+	ENGINE_UNSET = iota
```
dmacthedestroyer: I was originally thinking we'd make it `postgres` by default, but I think this is probably a better idea. Did you weight these two options?

types/types.override.pb.go
```diff
@@ -1,7 +1,30 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/golang/protobuf/jsonpb"
+)
+
+// MarshalJSONPB overloads UUIDValue's standard PB -> JSON conversion
+func (m *UUIDValue) MarshalJSONPB(*jsonpb.Marshaler) ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%s"`, m.Value)), nil
```
dmacthedestroyer: does this have to be `Sprintf`'d? If so, how come?

example/feature_demo/marshal_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestUnmarshalTypes(t *testing.T) {
+	ts, _ := ptypes.TimestampProto(time.Date(2009, 11, 17, 20, 34, 58, 651387237, time.UTC))
+	marshaller := &runtime.JSONPb{OrigName: true, EmitDefaults: true}
+	for in, expected := range map[string]TestTypes{
+		`{}`: {},
+		`{"api_only_string":"important text"}`:            {ApiOnlyString: "important text"},
+		`{"numbers":[1,2,3,4]}`:                           {Numbers: []int32{1, 2, 3, 4}},
+		`{"optional_string":"something real"}`:            {OptionalString: &wrappers.StringValue{Value: "something real"}},
+		`{"becomes_int":"GOOD"}`:                          {BecomesInt: 1},
+		`{"uuid":"123abc-unchecked-uuid"}`:                {Uuid: &types.UUIDValue{Value: "123abc-unchecked-uuid"}},
```
dmacthedestroyer: looks like you're trimming whitespace -- you should add tests with leading and trailing whitespace (if trimming it is even a good idea, which I'm not sure)

types/types.override.pb.go
```diff
@@ -1,7 +1,30 @@
...
+
+// UnmarshalJSONPB overloads UUIDValue's standard JSON -> PB conversion
+func (m *UUIDValue) UnmarshalJSONPB(_ *jsonpb.Unmarshaler, data []byte) error {
+	m.Value = strings.Trim(string(data), `"`)
+	return nil
+}
+
+// MarshalJSONPB overloads JSONValue's standard PB -> JSON conversion
+func (m *JSONValue) MarshalJSONPB(*jsonpb.Marshaler) ([]byte, error) {
+	return []byte(m.Value), nil
```
dmacthedestroyer: I know that you asked me about whether to validate this here or not and I said "not", but looking at it the way it is now, I'm thinking I was wrong.  In the big picture, what happens when these `Marshal`/`Unmarshal` functions return an error?

types/types.override.pb.go
```diff
@@ -1,7 +1,30 @@
 package types
 
-// XXX_WellKnownType is a hack -- please don't make a pattern out of this!
-// This is used to trick certain toolsets to treat this field like a WKT StringValue.
-// See https://github.com/golang/protobuf/blob/70c277a8a150a8e069492e6600926300405c2884/jsonpb/jsonpb.go#L157
-// and https://github.com/golang/protobuf/blob/70c277a8a150a8e069492e6600926300405c2884/jsonpb/jsonpb.go#L191
-func (*UUIDValue) XXX_WellKnownType() string { return "StringValue" }
+import (
```
dmacthedestroyer: I feel like you should have your `marshal_tests` here that relate to `UUID` and `JSONValue` instead of having them getting coincidentally tested in the `feature_demo` package. It could happen that somewhere down the line that `feature_demo` package gets removed and then we suddenly lose test coverage on these types.

example/feature_demo/marshal_test.go
Calebjh: I've seen the `Must` prefix around. I suppose I can create a func for this test to allow that to be inlined.

dmacthedestroyer: not important -- up to you.
It'd be useful if you were running different tests with multiple timestamps, which you're not doing. Just wanted to point out the pattern :)

types/types.override.pb.go
Calebjh: The Unmarshal strips double quotes and the Marshal re-adds them so that it isn't stored with the quotes. Sprintf just came to mind as the proper tool.

types/types.override.pb.go
Calebjh: Good call

types/types.override.pb.go
dmacthedestroyer: oops.. didn't notice the quotes in there. You're all good :)

plugin/plugin.go
Calebjh: I didn't really consider having `postgres` be default. Assuming something that specific sounds like a bad idea to me, especially since there hasn't been any documented mention of Postgres in this project (yet).

Makefile
Calebjh: Neither in general. However because the marshal test assume the JSON field exists they'll fail to build.

dmacthedestroyer: ah, interesting. I guess if we were trying to be pure, we'd have a proto run for each engine we know about (2 in this case) and tests or whatever for all of them

example/feature_demo/marshal_test.go
Calebjh: Are you referring to the strings.Trim? Because as above that's for the double-quotes. But I should add some tests for things like that, I agree.

dmacthedestroyer: ah, turns out I wasn't paying enough attention again. In this case, I'm assuming it'd also remove quotes from within the value, as well? `"some"really_bad_uuid_value"` -> `somereally_bad_uuid_value`

types/types.override.pb.go
johnbelamaric: %q instead of %s probably does what you want.

example/feature_demo/marshal_test.go
Calebjh: https://golang.org/pkg/strings/#Trim Just leading and trailing quotes. In theory though `"""super quoted string""""""` -> `super quoted string`

Calebjh: Should we trim whitespace? Whitespace isn't a valid part of UUIDs, is it?

dmacthedestroyer: my thought is we leave the input untouched -- so if they somehow put whitespace in it, we should preserve that. I didn't read your code properly (my fault) and thought that was what you were doing, but in fact you're just removing the quotes. Since we add the quotes in the first place, it makes sense to remove them too.

This is still a lazy way to do it because of the possibility of removing more than the leading and trailing quote. This is only an issue if we in fact don't validate the UUID during the un/marshal process.

types/types.override.pb.go
Calebjh: I don't think we need to validate here, because the overall JSON is already validated for syntax before being parsed into the individual fields. I'm adding some negative tests that will demonstrate this.

dmacthedestroyer: what validates it? the gRPC server?

Calebjh: https://github.com/golang/protobuf/blob/3a3da3a4e26776cc22a79ef46d5d58477532dede/jsonpb/jsonpb.go#L653-L670 The Unmarshaler uses the standard golang encoding/json to decode the JSON before converting the raw JSON to Protobuf.

PR #56, opened 2018-05-25
ManyToMany

dkukharau:


plugin/associations.go
```diff
@@ -201,6 +222,66 @@ func (p *OrmPlugin) parseBelongsTo(msg *generator.Descriptor, child *OrmableType
...
+	mtm.AssociationForeignkey = &assocKeyName
+	if foreignKey.Type != assocKey.Type {
+		p.Fail("Type mismatch of foreignkey and association foreignkey in", ormable.Name, fieldName, "association.")
+	}
+	var jt string
+	if jt = generator.CamelCase(mtm.GetJointable()); jt == "" {
+		if p.countManyToManyAssociationDimension(msg, fieldType) == 1 && typeName != fieldType {
+			jt = jgorm.ToDBName(typeName + fieldType)
+		} else {
+			jt = jgorm.ToDBName(typeName + inflection.Singular(fieldName))
```
Calebjh: Is the `Singular` here more intuitive? That results in `user_language` instead of `user_languages`, right? The gorm example http://gorm.io/docs/many_to_many.html has the plural instead. I'm open to either, but was wondering about the reasoning.

dkukharau: `user_language` appeared more natural for me on first sight, but it seems that everybody [suggests](https://stackoverflow.com/questions/1764483/sql-join-table-naming-convention) `user_languages` instead. Let's do it this way.

PR #60, opened 2018-05-31
Don't generate invalid code if child is not multi-account but parent is

Calebjh:
Code generated for removing child objects was based on whether parent was a multi-account type. It may be good to enforce children of multi-account objects also being multi-account, but since that isn't enforced currently, this will prevent invalid code from being generated.


Evgeniy-L:
@Calebjh, does this mean that we force a child to have account_id if parent has it? What if child object is only valid in scope of parent (e.g. cannot be controlled without parent). In this case "account_id" in child seems redundant.

Calebjh:
This was to fix a case where it checked the parent object instead of the child object for multi-account support before generating code that assumed the child had an account_id. It shouldn't be assumed that the child object is multi-account, and this aimed to fix a code-bug that caused it to behave as if that were required. This change is incorporated into PR #61 though, so I'm closing this.

PR #63, opened 2018-06-05
Removed range by sortFieldNames, Add range by descriptor and checking…

serkulesh:
Wrote a solution for creating the same structure fields of type as in the .proto file, and not sorting fields by ABC.

### DIFF
--- a/plugin/orderfield.go
+++ b/plugin/orderfield.go
@@ -4,18 +4,20 @@ import (
 	"github.com/gogo/protobuf/protoc-gen-gogo/generator"
 )
 
-type OrmOrderField struct{}
+func getOrderedFieldNames(ormable *OrmableType, message *generator.Descriptor) (fields []string) {
 
-func (o *OrmOrderField) getOrderedFieldNames(ormable *OrmableType, message *generator.Descriptor) (fields []string) {
-
+	// Iterate message fields
 	for _, v := range message.GetField() {
 		fieldName := generator.CamelCase(*v.Name)
-		fields = append(fields, fieldName)
+		_, ok := ormable.Fields[fieldName]
+		if ok {
+			fields = append(fields, fieldName)
+		}
 	}
 
-	ormFields := o.getOrmableFieldNames(ormable.Fields)
-	for _, ormField := range ormFields {
-		if !o.searchField(fields, ormField) {
+	// Iterate ormableType fields (include & fields)
+	for ormField := range ormable.Fields {
+		if !searchField(fields, ormField) {
 			fields = append(fields, ormField)
 		}
 	}
@@ -23,18 +25,11 @@ func (o *OrmOrderField) getOrderedFieldNames(ormable *OrmableType, message *gene
 	return fields
 }
 
-func (o *OrmOrderField) searchField(fields []string, searchName string) bool {
+func searchField(fields []string, searchName string) bool {
 	for _, fieldName := range fields {
 		if searchName == fieldName {
 			return true
 		}
 	}
 	return false
-}
-
-func (o *OrmOrderField) getOrmableFieldNames(fields map[string]*Field) (keys []string) {
-	for k := range fields {
-		keys = append(keys, k)
-	}
-	return keys
 }
--- a/plugin/orderfield.go
+++ b/plugin/orderfield.go
@@ -4,9 +4,10 @@ import (
 	"github.com/gogo/protobuf/protoc-gen-gogo/generator"
 )
 
+// Return "orm fields" with structure as in message
 func getOrderedFieldNames(ormable *OrmableType, message *generator.Descriptor) (fields []string) {
 
-	// Iterate message fields
+	// Iterate message fields and filtering by orm fields
 	for _, v := range message.GetField() {
 		fieldName := generator.CamelCase(*v.Name)
 		_, ok := ormable.Fields[fieldName]
@@ -15,7 +16,7 @@ func getOrderedFieldNames(ormable *OrmableType, message *generator.Descriptor) (
 		}
 	}
 
-	// Iterate ormableType fields (include & fields)
+	// Iterate ormableType fields, add (include & fields)
 	for ormField := range ormable.Fields {
 		if !searchField(fields, ormField) {
 			fields = append(fields, ormField)
--- a/plugin/orderfield.go
+++ b/plugin/orderfield.go
@@ -1,0 +1,40 @@
+package plugin
+
+import (
+	"github.com/gogo/protobuf/protoc-gen-gogo/generator"
+)
+
+type OrmOrderField struct{}
+
+func (o *OrmOrderField) getOrderedFieldNames(ormable *OrmableType, message *generator.Descriptor) (fields []string) {
+
+	for _, v := range message.GetField() {
+		fieldName := generator.CamelCase(*v.Name)
+		fields = append(fields, fieldName)
+	}
+
+	ormFields := o.getOrmableFieldNames(ormable.Fields)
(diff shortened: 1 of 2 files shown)


Calebjh:
Thank you for your interest in our project! 

I would like to see the fields in the generated object in the same order as the .proto file, and we used to use this exact approach, however we switched away because we may add extra fields not included in the Message object for an account_id or a foreign key, and needed a deterministic way to order the fields.

We could return to this for the fields present in the message and then append any added fields, but I think this would reduce overall code readability without any substantial benefits.

@dkukharau specifically made this change, I'll let him comment if he has any other input

serkulesh:
Thank you for the answer!. Hmm..Maybe I wrong but, sql table structure more readable as {id, user_id, firstname, lastname, created_at} than {created_at, firstname, id, lastname, user_id}. What do you think?

dkukharau:
@serkulesh Thanks for your PR! 
As @Calebjh mentioned, ormable may contain fields which are not part of proto message(foreign keys, accound id and fields included by `include` proto option), therefore your code doesn't generate correct GORM model.
I'm ok to maintain the ordering of fields defined in proto and append GORM specific fields to the end, but the ordering logic should be encapsulated in a function like `getSortedFieldNames` which returns a sequence of field names with correct order, so it can be used in different parts of generator, not only in `generateOrmable`.

serkulesh:
Thank you for the answer! I got it, and fixed include option in next commit (66a4f17). Check it please. 
Note: I merge fields of descriptor and ormPlugin fields, "include" fields added after descriptor fields.

plugin/plugin.go
```diff
@@ -216,12 +216,35 @@ func (p *OrmPlugin) getSortedFieldNames(fields map[string]*Field) []string {
...
+
+	for _, v := range message.GetField() {
+		fieldName := generator.CamelCase(*v.Name)
+		fields = append(fields, fieldName)
+	}
+
+	ormFields := p.getSortedFieldNames(ormable.Fields)
+	fieldsJoin := strings.Join(fields, " ")
+	for _, s := range ormFields {
+		if !strings.Contains(fieldsJoin, s) {
```
dkukharau: This check doesn't return correct answer if there is e.g. `ProfileId` field in PB message and `Id` in ORM message but not in PB.
Also take in consideration that some fields can conversely exist in PB but not in ORM, this case is not handled in your function.

serkulesh: @dkukharau hi, can you send me an example proto file, please? Thank you for the answers! :+1:

plugin/plugin.go
```diff
@@ -216,12 +216,35 @@ func (p *OrmPlugin) getSortedFieldNames(fields map[string]*Field) []string {
...
 
+func (p *OrmPlugin) generateFields(ormable *OrmableType, message *generator.Descriptor) []string {
+	var fields []string
+
+	for _, v := range message.GetField() {
+		fieldName := generator.CamelCase(*v.Name)
+		fields = append(fields, fieldName)
+	}
+
+	ormFields := p.getSortedFieldNames(ormable.Fields)
```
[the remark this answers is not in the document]
serkulesh: Yep sure!

plugin/plugin.go
dkukharau: @serkulesh Do you need any specific example? We have a bunch of them in `example` folder.

serkulesh: @dkukharau I think some specific example. I generated file from `example` folder, and compare them with files wich generated by my fork `protoc-gen-gorm`  and don't see differents. If you can send me some one specific proto file, i would be glad. Thank you!

dkukharau: Ok.
Try to insert `include: [{type: "int32", name: "id"}]` into `(gorm.opts)` of `TestTypes` from `/example/feature_demo/demo_types.proto` and see that this `id` field will not be generated in `TestTypesORM` when using your fork.
Regarding my second note(some fields can conversely exist in PB but not in ORM), I see that it's not actually true and your code handles correctly this case, but what I expect is that `getOrderedFieldNames` function should only return field names which exist in ORM, so it can be safely used across the plugin. I'm sorry for being inaccurate in my initial review:)

serkulesh: @dkukharau your true! My fault, my function has an error in comparing fields. I fixed compare, and created separate file for `getOrderedFieldNames`. Check it please :) Thank you for the example :)

plugin/orderfield.go
```diff
@@ -0,0 +1,40 @@
+package plugin
+
+import (
+	"github.com/gogo/protobuf/protoc-gen-gogo/generator"
+)
+
+type OrmOrderField struct{}
+
+func (o *OrmOrderField) getOrderedFieldNames(ormable *OrmableType, message *generator.Descriptor) (fields []string) {
```
dkukharau: As I stated in my previous comment, `getOrderedFieldNames` function should only return field names which exist in ORM, so it can be safely used across the plugin. So before appending a field to `fields` you should check if it exists in ORM model.

plugin/orderfield.go
```diff
@@ -0,0 +1,40 @@
+package plugin
+
+import (
+	"github.com/gogo/protobuf/protoc-gen-gogo/generator"
+)
+
+type OrmOrderField struct{}
```
dkukharau: Any reason to have this struct? I would just have these functions to be methods of `*OrmPlugin`, like everywhere in the plugin.

plugin/orderfield.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (o *OrmOrderField) getOrderedFieldNames(ormable *OrmableType, message *generator.Descriptor) (fields []string) {
+
+	for _, v := range message.GetField() {
+		fieldName := generator.CamelCase(*v.Name)
+		fields = append(fields, fieldName)
+	}
+
+	ormFields := o.getOrmableFieldNames(ormable.Fields)
```
dkukharau: Why not just iterate over `ormable.Fields`? Why need `getOrmableFieldNames` function?
```golang
for ormField := range ormable.Fields {
    ...
}
```

plugin/orderfield.go
serkulesh: Hi @dkukharau , About it `So before appending a field to fields you should check if it exists in ORM model.` - I checking field ORM in generateOrmable func, line 226. If need checking in my function i will do it. What do you think? :)
```
field, ok := ormable.Fields[fieldName]
if ok {
	p.P(fieldName, ` `, field.Type, p.renderGormTag(field))
}
```

plugin/orderfield.go
serkulesh: Hm..I will remove this struct.

plugin/orderfield.go
serkulesh: @dkukharau thanks. Will do iterate over `ormable.Fields`

plugin/orderfield.go
dkukharau: Yeah, I see that.
But as I stated in my previous comments:
`what I expect is that getOrderedFieldNames function should only return field names which exist in ORM, so it can be safely used across the plugin.`
I want this function to be used instead of `getSortedFieldNames`, and that function was supposed to return only ormable fields.
Thanks!

serkulesh:
Hi @dkukharau , about it `I expect is that getOrderedFieldNames function should only return field names which exist in ORM`. Added checking on orm fileds, and return orm fields with include fields. Can you check it please? Thank you.

dkukharau:
Hey @serkulesh 
Do you need some help with you PR? Please let us know if you think something is wrong. Thanks.

PR #64, opened 2018-06-05
Add default PK ordering

ystankevich:


plugin/handlergen.go
```diff
@@ -172,6 +172,16 @@ func (p *OrmPlugin) generateListHandler(message *generator.Descriptor) {
...
 	p.sortOrderedHasMany(message)
 	p.generatePreloading(ormable)
+
+	// add default ordering by primary key
+	pkName, pk := p.findPrimaryKey(ormable)
+	column := pk.GetTag().GetColumn()
+	if len(column) > 0 && column != pkName {
+		p.P(`db = db.Order("`, column, `")`)
+	} else {
+		p.P(`db = db.Order(gorm.ToDBName("`, pkName, `"))`)
```
Calebjh: I think `gorm.ToDBName(` should be executed in code-gen instead of part of the printed string to be evaluated at runtime.

plugin/handlergen.go
```diff
@@ -172,6 +172,16 @@ func (p *OrmPlugin) generateListHandler(message *generator.Descriptor) {
 	p.P(`db = db.Where(&ormParams)`)
 	p.sortOrderedHasMany(message)
 	p.generatePreloading(ormable)
+
+	// add default ordering by primary key
+	pkName, pk := p.findPrimaryKey(ormable)
+	column := pk.GetTag().GetColumn()
+	if len(column) > 0 && column != pkName {
```
johnbelamaric: What's not clear to me is what happens if the user passes in an _order_by? If they pass in no _order_by, we should sort by PK. If they pass in, say, "a, b" as the order_by, then we should sort by "a, b, PK". Will this do that?

ystankevich: In line 163 (called ApplyCollectionOperators), which handle request parameters and apply _oreder_by values  from user. So in line 180, 182 we just add additional PK ordering, which will be the latest.

PR #70, opened 2018-06-07
Add service level option to integrate with toolkit txn middleware

Calebjh:
It was noted that generated code does not integrate at all with transaction middleware offered by our atlas-app-toolkit.

- Adds a service option `(gorm.server).txn_middleware` which will cause the autogenerated server to (theoretically) use the transaction middleware instead of a server-wide DB instance.
- Renames `make build` to `make options` because that makes way more sense
- Cleaned up a couple import and package handling things
- Adds example service `IntPointTxn` that uses the `txn_middleware` option, but is otherwise identical to service `IntPoint` to demonstrate the difference in DB handling. (@amaskalenka please review this generated code to see if it uses the middleware properly).


example/feature_demo/demo_service.pb.gorm.go
```diff
@@ -302,3 +307,130 @@ func (m *IntPointDefaultServer) CreateSomething(ctx context.Context, in *Somethi
...
+// Create ...
+func (m *IntPointTxnDefaultServer) Create(ctx context.Context, in *CreateIntPointRequest) (*CreateIntPointResponse, error) {
+	if custom, ok := interface{}(m).(IntPointTxnCreateCustomHandler); ok {
+		return custom.CustomCreate(ctx, in)
+	}
+	txn, ok := tkgorm.FromContext(ctx)
+	if !ok {
+		return &CreateIntPointResponse{}, errors.New("Database Transaction For Request Missing")
+	}
+	db := txn.Begin()
```
Calebjh: I mimicked the behavior in the [Toolkit Readme ](https://github.com/infobloxopen/atlas-app-toolkit#middlewares). Did I miss anything important? It's hard to apply it to the example app without merging it so the gentool can grab it first.

dkukharau: I can tell you how I personally test my local changes to protoc-gen-gorm:
- Push your changes to your protoc-gen-gorm fork, e.g. `https://github.com/calebjh/protoc-gen-gorm`
- Clone the gentool
- Modify gentool's glide.yaml so it grabs protoc-gen-gorm from your fork:
```
  # gorm code generator
  - package: github.com/infobloxopen/protoc-gen-gorm
    repo: https://github.com/calebjh/protoc-gen-gorm
```
- Build gentool with the tag which is required by contacts-app(v3 now), modify it's makefile accordingly
- Generate contacts-app code
I personally don't feel any burden when doing this..

Calebjh:
I definitely agree we should carefully consider whether options belong in the proto file or in command line. In this case, I assumed it _might_ be the case that some services wanted to leverage TXN middleware while others might not. In which case it has to be service-level, not protoc level. Do you think is an okay assumption, @amaskalenka? I could also do both and add a protoc argument option that will apply to all services to allow for reduced proto decoration, and then either could be used depending on the desired application behavior.

example/feature_demo/demo_service.pb.gorm.go
Calebjh: That works great! I didn't know about the `repo` option in glide. Thanks!

PR #71, opened 2018-06-12
Enable cross-file and cross-package imported types

### LINKED ISSUE #68: Error on import message, is not ormable.
When import another proto file **import "color/color_service.proto"** and try to use message with "**option (gorm.opts) = {ormable: true}**", i got an error: "error:color.Color is not ormable.".

color_service.proto
```
message Color {
    option (gorm.opts) = {ormable: true};
    string name = 1;
}
```

part_service.proto
```
import "options/gorm.proto";
import "color/color_service.proto";
message Part {
    option (gorm.opts) = {ormable: true};
    string name = 1;
    int32 status = 2;
    color.Color color = 3;
}
```

Calebjh:
int_point -> IntPoint for best practices naming conventions

Fix handling of types imported from other .proto files. Note, typename uniqueness is still currently required (can't have packageA.typeOne and packageB.typeOne).

Resolves #68


dkukharau:
This is very ambitious change! Didn't know it's possible to do something in this area. If you don't mind I will look more deeply at this tomorrow.

dkukharau:
Well, this looks good for me even I'm still not getting all the package/import handling magic behind:), I need to dive into gen-go-go code more in order to understand all the bits.. Excellent work!
I want to ask one thing which is not very clear for me from my current understanding, but anyway: where is this limitation on cross-package has-one and has-many coming from? It seems for me that you generate all the files passed to protoc with a single instance of the plugin, so it can maintain state between file generation, and I don't see a difference if files are in same package or not in the context of associations. Sorry if I'm missing something obvious.

dkukharau:
Well, got it. Thanks!

PR #72, opened 2018-06-13
Fix position field handling

dkukharau:


Evgeniy-L:
LGTM

example/user/user.proto
```diff
@@ -80,4 +80,5 @@ message Task {
     };
     string name = 1;
     string description = 2;
+    int64 priority = 3;
```
Calebjh: Are you manually including the field here to demonstrate that it works properly with an `int64` type in addition to just the automatic `int`?

dkukharau: Yeah, exactly.

PR #74, opened 2018-06-15
Add support for including types of non-builtin types

Calebjh:
Changes for a fix and improvement.
- No longer generates broken code to use e.g. `time.Time` as an `include` field if a google.protobuf.Timestamp type is not used elsewhere in file.
- Can add arbitrary types with format `package_alias.TypeName` in include, as long as the `package: "path"` option is used.


plugin/plugin.go
```diff
@@ -213,13 +232,55 @@ func (p *OrmPlugin) parseBasicFields(msg *generator.Descriptor) {
...
+					fmt.Sprintf(`Included field %q of type %q reuses package prefix %q, but has different package %q, was %q`,
+						field.GetName(), field.GetType(), parts[0], field.GetPackage(), v))
+			}
+			p.GetFileImports().githubImports[parts[0]] = field.GetPackage()
+		} else if v, ok := p.GetFileImports().githubImports[parts[0]]; !ok {
+			p.Fail(fmt.Sprintf(`Included field %q of type %q not a recognized special type, and no package specified`, field.GetName(), field.GetType()))
+		} else {
+			// If this package prefix has already been used with a package definition,
+			// let's not give up, but let them know and hope for the best
+			log.Printf(`Included field %q of type %q reuses package prefix %q, assumed to also be from package %q`,
```
dkukharau: What is the need of specifying alias along with the type name? It even complicates things since you need to validate if the user did a mistake by duplicating aliases.

plugin/imports.go
```diff
@@ -33,58 +34,89 @@ func CleanImports(pFileText *string) *string {
...
-			githubImports["gtypes"] = "github.com/infobloxopen/protoc-gen-gorm/types"
-		}
-	}
 	for _, typeName := range imports.typesToRegister {
 		p.RecordTypeUse(typeName)
 	}
-	sort.Strings(stdImports)
-	for _, dep := range stdImports {
+	githubImports := imports.packages
+	sort.Strings(imports.stdImports)
```
dkukharau: `Gofmt` does sorting of imports which is called for each generated file. Is it not enough for us?

dkukharau:
Thanks for doing this refactoring. It looks much better now!
This doesn't require any addressing, I just want to get your understanding:
Why you treat differently standard/non-standard packages and predefine alias for a package? Do you want to have a possibility to hardcode alias in code-gen expressions or you just don't want to break existing ones?

Calebjh:
Mostly for the predefined aliases, statements like ``p.P(`gorm.DB`)`` instead of having to do ``p.P(p.AddImport("github.com/jinzhu/gorm"), `.DB`)`` 
It also has benefits like if the import path changes in the future we can change it one place instead of having to replace all occurrences.
I suppose if `AddImport` was always used then there would be no need to specify that the imports are needed elsewhere. Do you think that would be overall better to use the AddImport method everywhere, maybe with the known imports as const strings instead?

dkukharau:
We could define some global constants like `gormPkgPath`, `authPkgPath` etc.. so it becomes p.P(p.AddImport(gormPkgPath), `.DB`). It will become a bit shorter and will be able to modify import path in one place if it changes.

dkukharau:
I'm okay with the current state, just wanted to know your motivation.

plugin/imports.go
Calebjh: I haven't noticed that the imports are automatically sorted. 
I just removed the sort statements and rebuilt and got this:
```
import time "time"
import context "context"
import errors "errors"

import gorm2 "github.com/infobloxopen/atlas-app-toolkit/gorm"
import ptypes1 "github.com/golang/protobuf/ptypes"
import auth1 "github.com/infobloxopen/atlas-app-toolkit/auth"
import gorm1 "github.com/jinzhu/gorm"
```

Calebjh:
@dkukharau I was hesitant, but I made the change since I figured it would make it easier in the future. If you're okay with this, can I get an official 'Approve' review for the PR?

plugin/imports.go
dkukharau: oops, I'm sorry for misinformation.

dkukharau:
Yes, sure. Thanks, good work!

PR #76, opened 2018-06-19
Add support for Postgres Inet type as PB gorm.types.InetValue

Calebjh:
Another special type that could be useful.

Should merge PR #74 first so that the inet type can be included in the updated import code.


plugin/plugin.go
```diff
@@ -347,6 +351,19 @@ func (p *OrmPlugin) renderGormTag(field *Field) string {
 			res += fmt.Sprintf("association_jointable_foreignkey:%s;", mtm.GetAssociationJointableForeignkey())
 		}
 	}
+	if field.Type == "*gtypes.Inet" {
+		if p.dbEngine == ENGINE_POSTGRES {
+			if res == "" {
+				return "`sql:\"type=inet\"`"
+			}
+			return fmt.Sprintf("`gorm:\"%s\" sql:\"type=inet\"`", strings.TrimRight(res, ";"))
```
dkukharau: `sql` tag is just the ancestor of `gorm` tag, so I think it's more preferable just to set `FieldOptions` for the field in `parseBasicFields`. Probably we want the same behavior for `uuid`.

types/inet_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	_ "github.com/lib/pq"
+)
+
+func TestInetParse(t *testing.T) {
+	inet, err := ParseInet("1.2.3.4/32")
+	if err != nil {
+		t.Error(err)
+	}
+	if !inet.IP.Equal(net.ParseIP("1.2.3.4")) || !reflect.DeepEqual(inet.Mask, net.CIDRMask(32, 32)) {
```
dkukharau: This is just a suggestion, not necessary to change:
This checks can be more easily done with something like https://github.com/stretchr/testify. I especially love it because if you have complex structs with a lot of nested messages and the check is not passed, it will print you the diff between compared structs. This is very useful when something is suddenly broken and you try to figure what is wrong.
Just an example from the [toolkit](https://github.com/infobloxopen/atlas-app-toolkit/blob/master/gorm/filtering_test.go).

Calebjh:
The intention was that it would contain an IP value & Mask, which can represent either a subnet & mask (`192.168.0.0/16`) or an IP address & mask (`192.168.1.1/16`), depending on how you handle it. Does the inet/cidr distinction seem crucial to you? It seems to me like the function of cidr is a strict subset of inet.

dkukharau:
No, I just wanted to understand potential use case of this type. I got it now, both pg `inet` and Golang's `Inet` represent either some subnet or IP address in a subnet. That's fine. 
I wanted to clarify because N* will likely ask about a special type for just an IP Address(`192.168.1.1`). It's something that they had in previous Infoblox products and what likely will  have in new ones.

PR #78, opened 2018-06-20
Support of infoblox.rpc.Identifier

amaskalenka:


Calebjh:
The package import and alias handling will have to be altered to fit with the recent cleanup (sorry about that, I promise this way is better though).
Have the Identifiers been tested in the Contacts app yet?

plugin/plugin.go
```diff
@@ -42,14 +43,16 @@ var wellKnownTypes = map[string]string{
...
+	OriginName string
+	Name       string
+	Package    string
+	Fields     map[string]*Field
 }
 
 type Field struct {
 	Type string
 	*gorm.GormFieldOptions
+	BindTo string
```
dkukharau: Could you make it `Parent *OrmableType` so it can be used a bit more universally across the plugin?

plugin/plugin.go
```diff
@@ -114,6 +117,7 @@ func (p *OrmPlugin) Generate(file *generator.FileDescriptor) {
 				if getMessageOptions(msg).GetOrmable() {
 					ormable := NewOrmableType()
 					ormable.Package = fileProto.GetPackage()
+					ormable.OriginName = typeName
```
dkukharau: Could pass `OriginName` and `Package` to `NewOrmableType` so all fields are initialized in constructor, as it should be?

plugin/plugin.go
```diff
@@ -129,6 +133,11 @@ func (p *OrmPlugin) Generate(file *generator.FileDescriptor) {
 				typeName := generator.CamelCase(msg.GetName())
 				if p.isOrmable(typeName) {
 					p.parseAssociations(msg)
+					o := p.getOrmable(typeName)
+					if p.hasPrimaryKey(o) {
+						_, fd := p.findPrimaryKey(o)
+						fd.BindTo = o.OriginName
```
dkukharau: I guess we don't need this `Parent` information in a field if it's not a foreign key. 
In Converters you may check if `Parent` is not nil, then use `Parent`s type name, otherwise use current type name.

plugin/plugin.go
```diff
@@ -250,7 +262,12 @@ func (p *OrmPlugin) generateOrmable(message *generator.Descriptor) {
 	p.P(`type `, ormable.Name, ` struct {`)
 	for _, fieldName := range p.getSortedFieldNames(ormable.Fields) {
 		field := ormable.Fields[fieldName]
-		p.P(fieldName, ` `, field.Type, p.renderGormTag(field))
+		if field.BindTo != "" {
```
dkukharau: I guess this is debug information and should be removed.

plugin/plugin.go
```diff
@@ -389,7 +407,11 @@ func (p *OrmPlugin) generateConvertFunctions(message *generator.Descriptor) {
 		if getFieldOptions(field).GetDrop() {
 			continue
 		}
-		p.generateFieldConversion(message, field, true)
+		var ofield *Field
+		if ormable != nil {
```
dkukharau: I guess this check is redundant since there is always should be an `ormable` and `getOrmable` fails if it doesn't exist for some reason.

plugin/plugin.go
```diff
@@ -423,7 +445,11 @@ func (p *OrmPlugin) generateConvertFunctions(message *generator.Descriptor) {
 		if getFieldOptions(field).GetDrop() {
 			continue
 		}
-		p.generateFieldConversion(message, field, false)
+		var ofield *Field
+		if ormable != nil {
```
dkukharau: Same here.

plugin/plugin.go
amaskalenka: That is needed for primary as well to decode/encode self id based on right proto type. If I remove it I would need to add complexity during generation.
Now I know if `fd.BindTo == ""` it is an external identifier.

example/user/user.proto
```diff
@@ -27,40 +27,40 @@ message User {
...
         ormable: true,
         multi_account: true
     };
-    infoblox.rpc.Identifier id = 1;
+    atlas.rpc.Identifier id = 1 [(gorm.field).tag = {type: "integer"}];
     string address_1 = 2;
     string address_2 = 3;
     string post = 4;
-    infoblox.rpc.Identifier external = 5;
+    atlas.rpc.Identifier external = 5 [(gorm.field).tag = {type: "jsonb"}];
```
Evgeniy-L: why jsonb (not string)?

Gopkg.lock
```diff
@@ -90,16 +90,18 @@
   revision = "8e809c8a86450a29b90dcc9efbf062d0fe6d9746"
 
 [[projects]]
+  branch = "resource-amaskalenka"
```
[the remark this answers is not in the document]
amaskalenka: Sure

example/user/user.proto
```diff
@@ -28,35 +29,38 @@ message User {
...
 message Email {
     option (gorm.opts) = {
         ormable: true,
         multi_account: true
     };
-    int32 id = 1;
+    atlas.rpc.Identifier id = 1 [(gorm.field).tag = {type: "uuid"}];
     string email = 2;
     bool subscribed = 3;
+    atlas.rpc.Identifier user_id = 4;
```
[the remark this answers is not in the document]
amaskalenka: That is for demostrate how fk could be a pesos on API.

example/user/user.proto
```diff
@@ -28,35 +29,38 @@ message User {
...
     option (gorm.opts) = {
         ormable: true,
         multi_account: true
     };
-    int32 id = 1;
+    atlas.rpc.Identifier id = 1 [(gorm.field).tag = {type: "integer"}];
     string address_1 = 2;
     string address_2 = 3;
     string post = 4;
+    atlas.rpc.Identifier external = 5 [(gorm.field).tag = {type: "jsonb"}];
```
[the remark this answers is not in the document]
amaskalenka: It could be “csp/hosts/73” reference to an external resource.

If you really want to understand use case please read through N* docs.

PR #79, opened 2018-06-27
Support collection parameters for list operation

ystankevich:


example/feature_demo/demo_service.proto
```diff
@@ -67,6 +68,13 @@ message ListIntPointResponse {
 message Something {
 }
 
+message CollectionOps {
```
amaskalenka: It seems that it should be renamed it to `ListContactRequest`

plugin/handlergen.go
```diff
@@ -161,9 +161,9 @@ func (p *OrmPlugin) generateListHandler(message *generator.Descriptor) {
 
 	p.P(`// DefaultList`, typeName, ` executes a gorm list call`)
 	p.P(`func DefaultList`, typeName, `(ctx context.Context, db *`, p.Import(gormImport), ``,
-		`.DB) ([]*`, typeName, `, error) {`)
+		`.DB, params interface{}) ([]*`, typeName, `, error) {`)
 	p.P(`ormResponse := []`, ormable.Name, `{}`)
-	p.P(`db, err := `, p.Import(tkgormImport), `.ApplyCollectionOperators(db, ctx)`)
+	p.P(`db, err := `, p.Import(tkgormImport), `.ApplyCollectionOperators(db, params)`)
```
Calebjh: The function signature is changing in atlas-app-toolkit? 
And this requires that any/all operators be included in the request, I guess this makes sense, but has there been a consensus on this?

ystankevich: I also send for review changes in atlas-app-toolkit.

PR #81, opened 2018-07-07
PUT operation create object if it doesn't exist

ystankevich:
If put is called for non existing object, it will be created. 
To set valid response code when object created, add that changes.
Fix for NGP-927


dmacthedestroyer:
Is this the actual behavior we want to support? I think if you try to `PUT` to a resource that doesn't exist, you should get a 404.

PR #83, opened 2018-07-10
fix broken unit test

dmacthedestroyer:
I pulled from master and ran `make test` -- there was a failed unit test. There were also modifications to 3 other files, due to running protoc.

It looks like the failed test was supposed to return an error but was instead executing successfully. Is this functionality now supported?

### DIFF
--- a/example/feature_demo/marshal_test.go
+++ b/example/feature_demo/marshal_test.go
@@ -16,7 +16,7 @@ import (
 func TestSuccessfulUnmarshalTypes(t *testing.T) {
 	unmarshaler := &jsonpb.Unmarshaler{}
 	for in, expected := range map[string]TestTypes{
-		`{}`: {},
+		`{}`:                                              {},
 		`{"api_only_string":"important text"}`:            {ApiOnlyString: "important text"},
 		`{"numbers":null}`:                                {},
 		`{"numbers":[]}`:                                  {Numbers: []int32{}},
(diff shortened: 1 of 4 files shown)


example/feature_demo/marshal_test.go
```diff
@@ -61,22 +61,23 @@ func TestBrokenUnmarshalTypes(t *testing.T) {
 	unmarshaler := &jsonpb.Unmarshaler{}
 	for in, expected := range map[string]string{
 		// A subset of possible broken inputs
-		`{"}`: "unexpected EOF",
-		`{"becomes_int":"NOT_AN_ENUM_VALUE"}`:       "unknown value \"NOT_AN_ENUM_VALUE\" for enum example.TestTypesStatus",
-		`{"numbers":[1,2,3,4,]}`:                    "invalid character ']' looking for beginning of value",
-		`{"json_field":{"top":{"something":1},2]}}`: "invalid character '2' looking for beginning of object key string",
-		`{"type_with_id_id":"4"}`:                   "json: cannot unmarshal string into Go value of type uint32",
```
dmacthedestroyer: This is the test I removed because it wasn't returning an error anymore

Calebjh: It appears that this was changed (fixed) in: https://github.com/golang/protobuf/commit/9eb2c01ac278a5d89ce4b2be68fe4500955d8179#diff-2e0364a4cb34e49f1b55ffaf06d6e8b4

PR #85, opened 2018-07-10
Add GORM tags for relationships to field messages

dmacthedestroyer:
putting this out early to get more eyes on it and see if it's a good idea or not


plugin/plugin.go
```diff
@@ -316,10 +316,10 @@ func (p *OrmPlugin) addIncludedField(ormable *OrmableType, field *gorm.ExtraFiel
 		} else if rawType == "Inet" {
 			rawType = fmt.Sprintf("%s.Inet", p.Import(gtypesImport))
 		} else {
-			p.Fail(
-				fmt.Sprintf(
-					`Included field %q of type %q is not a recognized special type, and no package specified`,
-					field.GetName(), field.GetType()))
+			//p.Fail(
```
dmacthedestroyer: for the example I wrote, having this code fail was getting in the way. I'm assuming this error was in place to give a user-friendly error for an edge case. Is there a way we can make it so that you can reference a type that's in the same package?

example/feature_demo/demo_types.pb.gorm.go
```diff
@@ -174,16 +175,17 @@ type TestTypesWithAfterToPB interface {
...
-	SecretInt     int32           `gorm:"-"`
-	Things        []*TestTypesORM `gorm:"foreignkey:ThingsTypeWithIDId;association_foreignkey:Id"`
-	User          *user.UserORM   `gorm:"foreignkey:UserId;association_foreignkey:Id"`
-	UserId        *string
+	ANestedObject     *TestTypesORM `gorm:"foreignkey:ANestedObjectTypeWithIDId;association_foreignkey:Id"`
+	Address           *types1.Inet  `gorm:"type:inet"`
+	Id                uint32
+	IntPointId        *uint32
+	Ip                string `gorm:"column:ip_addr"`
+	MultiAccountTypes []*JoinTable
```
dmacthedestroyer: TODO: find out why the GORM tags weren't added to this

plugin/plugin.go
Calebjh: This is why I said you'd probably have to put the `JoinTable` type into a separate package (then using the `package: "..."` tag. Otherwise there's this.

dmacthedestroyer: do we need to fail on this? or can we add a special value or something to the `package:` option to indicate that it's from the same package? What do you think would be best?

example/feature_demo/demo_types.proto
```diff
@@ -70,7 +70,8 @@ message TypeWithID {
   option (gorm.opts) = {
     ormable: true,
     include: [
-      {type: "int32", name: "secret_int", tag: {ignore: true}}
+      {type: "int32", name: "secret_int", tag: {ignore: true}},
+      {type: "[]*JoinTable", name: "multi_account_types", tag: {foreign_key: "TypeWithIDID"}}
```
dmacthedestroyer: This is my particular use case for this PR: being able to specify a GORM model that shouldn't be visible to the API spec, but still available to the GORM models

dmacthedestroyer:
This looks ready to go to me... please let me know if you see any red flags with allowing this functionality.

PR #89, opened 2018-07-13
Support of implicit foreign key of type atlas.rpc.Identifier.

amaskalenka:
The intention of added `reference_of` option is to allow calculate resource type of `atlas.rpc.Identifier` properly in case when an identifier is implicit foreign key.

NOTE: The **implicit foreign key** is a foreign key defined on DB level via constraints but in Proto file it is defined as simple field that holds id of some object without any association.


Evgeniy-L:
looks ok to me

example/user/user.proto
```diff
@@ -55,6 +55,7 @@ message Address {
     string address_2 = 3;
     string post = 4;
     atlas.rpc.Identifier external = 5 [(gorm.field).tag = {type: "jsonb"}];
+    atlas.rpc.Identifier implicit_fk = 6 [(gorm.field) = {reference_of: "Email" tag: {type: "text"} }];
```
Calebjh: Is there not a comma needed between `"Email" tag:`?

amaskalenka: The comma is not required

amaskalenka:
Exactly! That is why I call it implicit foreign key.

PR #90, opened 2018-07-13
Switch function name in error to match properly

### LINKED ISSUE #62: Default Strict Update function refers to itself with wrong name in error
e.g. 
```
return nil, fmt.Errorf("Nil argument to DefaultCascadedUpdateContact")
```
for function `DefaultStrictUpdateContact`

Calebjh:
5-minute fix to cleanup an old issue

Resolves #62


wsp93:
I've taken a look at the code and the corresponding issue, and the code seems to fix the corresponding issue. LGTM :)

PR #92, opened 2018-07-16
Prehook to replace fairly useless method override hook

Calebjh:
Instead of overriding entire autogenerated functions via hook, approach is now to embed the autogenerated server in a new object and override only the necessary functions (also grants access to the autogenerated method).
e.g. 
```
type CustomizedServer {
    GeneratedServer
}
func (cs *CustomizedServer) Create(...) {
    // Do special stuff
    cs.GeneratedServer.Create(...) // if desired, or can override all functionality
}
```

Prehook can be used for altering values in context, adding logging, setting values within the message, etc.


Calebjh:
Well, I can't fix that, but a week from now if it'll still be helpful here it is.

PR #94, opened 2018-07-16
if inet mask refers to single host, don't print it

dmacthedestroyer:


types/inet_test.go
```diff
@@ -49,3 +49,28 @@ func TestInetParse(t *testing.T) {
...
+		name  string
+		input string
+		want  string
+	}{
+		{"v4", "192.168.1.1", "192.168.1.1"},
+		{"v4 with single host mask", "192.168.1.1/32", "192.168.1.1"},
+		{"v4 with different mask", "192.168.1.1/24", "192.168.1.1/24"},
+		{"v6", "2001:0db8:85a3:0000:0000:8a2e:0370:7334", "2001:db8:85a3::8a2e:370:7334"},
+		{"v6 with single host mask", "2001:0db8:85a3:0000:0000:8a2e:0370:7334/128", "2001:db8:85a3::8a2e:370:7334"},
+		{"v6 with different", "2001:0db8:85a3:0000:0000:8a2e:0370:7334/64", "2001:db8:85a3::8a2e:370:7334/64"},
```
kd7lxl: `Inet.IP.String` doesn't collapse the consecutive zeros?

kd7lxl: Oh, I totally missed that you already had that in there.

PR #100, opened 2018-07-24
Modify default preloading behavior

dkukharau:


plugin/handlergen.go
```diff
@@ -210,6 +213,9 @@ func (p *OrmPlugin) generateListHandler(message *generator.Descriptor) {
 	p.P(`if err != nil {`)
 	p.P(`return nil, err`)
 	p.P(`}`)
+	p.P(`if fs.GetFields() == nil {`)
```
Calebjh: Is this checked for validity anywhere? Don't we need to make sure the Request type has a `fields` field before generating this code?

dkukharau: No, we don't need it. `fs` is extracted in runtime with a bunch of reflect stuff. We could probably make this collection operators handling more statically typed in future, but that's what we have for now. 
See the [PR](https://github.com/infobloxopen/atlas-contacts-app/pull/57) in contacts-app.

Calebjh: Ah. I see now, I was thinking that was being called on the request object directly.

PR #107, opened 2018-08-01
Fix empty import / nil reference handling

dkukharau:
- Package wasn't set for included fields causing empty imports.
- Don't try to type-assert nil driver.Value


example/user/user.pb.gorm.go
```diff
@@ -169,7 +169,7 @@ func (m *User) ToORM(ctx context.Context) (UserORM, error) {
 	if m.ExternalUuid != nil {
 		if v, err := resource1.Decode(nil, m.ExternalUuid); err != nil {
 			return to, err
-		} else {
+		} else if v != nil {
 			vv := v.(string)
```
Calebjh: The `nil` check is good, but should we do a check on this type assertion, too? Or are we okay with, panic-ing if the DB type is wrong (what would cause this assert to fail, I believe) since that's a big error, or should we do something more "graceful"?

PR #108, opened 2018-08-03
Make PATCH recursively apply Fieldmask

Calebjh:
Previous PATCH handling code won't handle nested objects properly (the subfields of the object are supposed to be updatable, and the path directly to an object will never be included if the FieldMask is set to spec). This change will cause the field mask to be recursively applied for these fields when present in the FieldMask.


Calebjh:
Ah. Should probably add a nil pointer check or two in places.

plugin/handlergen.go
```diff
@@ -118,23 +118,71 @@ func (p *OrmPlugin) generateReadHandler(message *generator.Descriptor) {
...
-		p.P(`if f == "`, ccName, `" {`)
-		p.P(`patchee.`, ccName, ` = patcher.`, ccName)
-		p.removeChildAssociationsByName(message, ccName)
-		p.setupOrderedHasManyByName(message, ccName)
-		p.P(`}`)
+		fieldType, _ := p.GoType(message, field)
+		//  for ormable message, do recursive patching
+		if field.IsMessage() && p.isOrmable(fieldType) && !field.IsRepeated() {
+			p.UsingGoImports("strings")
+			p.P(`if strings.HasPrefix(f, prefix+"`, ccName, `.") && !updated`, ccName, ` {`)
```
dmacthedestroyer: say we have the following:
```
struct {
  Foo struct {
    Bar struct {
      Baz string
    }
  }
}
```
and FieldMask with path `Foo`

Will this catch it? I see the dot at the end which would imply that a full replacement of `Foo` would get missed

Calebjh: Ah. You are correct.

PR #111, opened 2018-08-10
Replace db.Save with StrictUpdate

askurydzin:
* Introduces a breaking changes: signature of the DefaultApplyFieldMask changed to exclude ormObj.

 * DefaultApplyFieldMask should not remove associations
 * Instead, DefaultStrictUpdate should be called after FieldMask application.


Gopkg.toml
```diff
@@ -8,7 +8,7 @@
 
 [[constraint]]
   name = "github.com/grpc-ecosystem/grpc-gateway"
-  version = "1.3.1"
+  version = "1.4.1"
```
Calebjh: How much of a difference does this make?

plugin/handlergen.go
```diff
@@ -270,29 +259,8 @@ func (p *OrmPlugin) generatePatchHandler(message *generator.Descriptor) {
...
-	p.P(`return nil, err`)
-	p.P(`}`)
-
-	// convert ormObj to pbObj again (sic!) to trigger any logic that was
-	// written for AfterToPB/BeforeToPB.
-	p.P(`pbObj, err = ormObj.ToPB(ctx)`)
-	p.P(`if err != nil {`)
-	p.P(`return nil, err`)
-	p.P(`}`)
+	p.P(`return DefaultStrictUpdate`, typeName, `(ctx, &pbObj, db)`)
```
Calebjh: Definitely cleaner (and also fixes the issue of removing associations in the ApplyFieldMask method). Have you tested it in practice to make sure it works correctly like this?

askurydzin: build passes, and only thing that has changed is apply field mask is now not removes associations (the're all are removed in StrictUpdate)

Gopkg.toml
askurydzin: it varies from repo to repo, however I would make it consistent with gentool

askurydzin:
noticed, that ormObj is no longer needed.

PR #113, opened 2018-08-13
Sustainable examples generation via gentool

askurydzin:
To avoid constant overriding of generated files I propose using gentool as a source container for building a temprorary protoc-gen-grom-gentool container that builds new protoc-gen-gorm and replaces the old one for generation step.

I would recommend to use `make gentool-example` at least once right before commit, while using existing `make example` for dev purposes.


askurydzin:
+ contacts-app can be updated using following command `make protobuf DOCKER_GENERATOR=infoblox/atlas-gentool-gengorm:latest`

Evgeniy-L:
how about `infoblox/atlas-gentool:dev-gengorm` as an image name?

PR #114, opened 2018-08-13
 Fix option names inconsistency

EvgenyMilev:


options/gorm.proto
```diff
@@ -68,7 +68,7 @@ message GormTag {
     optional string embedded_prefix = 13;
     optional bool ignore = 14;
     optional string foreign_key = 15;
-    optional string association_foreign_key = 16;
+    optional string association_foreignkey = 16;
     optional string many_to_many = 17;
     optional string jointable_foreignkey = 18;
     optional string association_jointable_foreignkey = 19;
```
Evgeniy-L: From my perspective `foreign_key` looks better than `foreignkey`, but I see that second variant is more frequent in proto. What about line 70?
I believe this is a "breaking change"?

dkukharau: `foreignkey` was was the initial choice in order to be consistent with [gorm](http://gorm.io/docs/models.html#Struct-tags-for-Associations)

PR #117, opened 2018-08-16
Added support for advanced association options

richardfelkl:
- For each association type new options have been added (association_autoupdate, association_autocreate, association_save_reference).
- Those options can be set to boolean value (by default set to `true` in GORM) represented by string which is parsed and checked in plugin/associations.go.
- Updated readme now mentions those options and contains link to official docs how to use them.


dkukharau:
Thanks, good work!

PR #120, opened 2018-08-22
Major revisit of hooks

dkukharau:
Summary:
- Remove outdated DefaultUpdate handler
- Add handler level hooks
- Add after service hooks

:bomb:**Breaking changes:**
- Fix BeforePatchSave hook signature
Reasoning: see a huge anti-pattern in a hook having `Context` as a return value.
- Fix before service hook signature
- Change DefaultList handler signature go receive collection operators similar to DefaultRead(#109)


example/feature_demo/demo_service.pb.gorm.go
```diff
@@ -103,52 +103,82 @@ func DefaultCreateIntPoint(ctx context.Context, in *IntPoint, db *gorm1.DB) (*In
 	if err != nil {
 		return nil, err
 	}
+	if hook, ok := interface{}(&ormObj).(IntPointORMWithBeforeCreate); ok {
+		db, err = hook.BeforeCreate(ctx, db)
+		if err != nil {
```
Calebjh: I see a lot of places using this form instead of the more condensed `if {db,} err = ...; err != nil {` form. Is there a reason behind that, or is it just coincidental?

plugin/servergen.go
```diff
@@ -420,28 +454,61 @@ func (p *OrmPlugin) getMethodProps(method *descriptor.MethodDescriptorProto) (ge
...
+func (p *OrmPlugin) hasSorting(object generator.Object) string {
+	return p.hasFieldOfType(object, "Sorting")
+}
+
+func (p *OrmPlugin) hasPagination(object generator.Object) string {
+	return p.hasFieldOfType(object, "Pagination")
 }
 
-func (p *OrmPlugin) hasFieldsSelector(object generator.Object) string {
+func (p *OrmPlugin) hasFieldOfType(object generator.Object, fieldType string) string {
```
Calebjh: I'm the one who first made this naming choice `has...`, but since it returns the field name, `get...` probably makes more sense.

example/feature_demo/demo_service.pb.gorm.go
dkukharau: I don't know.. Perhaps I like this way because it's quite easy to copy-paste around.. But honestly I don't care much about this..

dkukharau:
Just added a test for this.

PR #123, opened 2018-09-10
Include vendored dependency instructions in prerequisites.

jaytaylor:
Include vendored dependency instructions in prerequisites.

Also cleaned up links and markdown header nesting.

---
Just want to also say a big thanks for creating this project - very, very cool!!!


Calebjh:
Thank you for your interest and contribution!
It was built as an internal solution, but I'm trying to keep things modularized enough to be publicly practical, I hope you find it useful.

PR #127, opened 2018-09-24
added changes in the generator for paging  improvements

EvgenyMilev:


example/feature_demo/demo_service.pb.gorm.go
```diff
@@ -684,10 +684,30 @@ func (m *IntPointServiceDefaultServer) List(ctx context.Context, in *ListIntPoin
 			return nil, err
 		}
 	}
+	pagedRequest := false
+	if in.Paging.Limit >= 1 {
```
Evgeniy-L: what will happen if request has no pagging?

PR #132, opened 2018-10-09
Fix Update concurrency issues

dkukharau:
- Reorder queries in Update handler(first update main object row so lock is acquired for this row and all concurrent Update transactions are waiting for the current transaction to finish, then cleanup/insert all associated objects).
- Refactor AccountId handling in List and Update

### DIFF
--- a/example/feature_demo/demo_service.pb.gorm.go
+++ b/example/feature_demo/demo_service.pb.gorm.go
@@ -349,21 +349,16 @@ func DefaultStrictUpdateIntPoint(ctx context.Context, in *IntPoint, db *gorm1.DB
 	if err != nil {
 		return nil, err
 	}
-	count := 1
-	err = db.Model(&ormObj).Where("id=?", ormObj.Id).Count(&count).Error
-	if err != nil {
-		return nil, err
-	}
-	if hook, ok := interface{}(&ormObj).(IntPointORMWithBeforeStrictUpdateSave); ok {
-		if db, err = hook.BeforeStrictUpdateSave(ctx, db); err != nil {
+	var count int64
+	lockedRow := &IntPointORM{}
+	count = db.Model(&ormObj).Set("gorm:query_option", "FOR UPDATE").Where("id=?", ormObj.Id).First(lockedRow).RowsAffected
+	if hook, ok := interface{}(&ormObj).(IntPointORMWithBeforeStrictUpdateCleanup); ok {
+		if db, err = hook.BeforeStrictUpdateCleanup(ctx, db); err != nil {
 			return nil, err
 		}
 	}
-	if err = db.Set("gorm:association_autoupdate", false).Set("gorm:association_autocreate", false).Set("gorm:association_save_reference", false).Save(&ormObj).Error; err != nil {
-		return nil, err
-	}
-	if hook, ok := interface{}(&ormObj).(IntPointORMWithBeforeStrictUpdateCleanup); ok {
-		if db, err = hook.BeforeStrictUpdateCleanup(ctx, db); err != nil {
+	if hook, ok := interface{}(&ormObj).(IntPointORMWithBeforeStrictUpdateSave); ok {
+		if db, err = hook.BeforeStrictUpdateSave(ctx, db); err != nil {
 			return nil, err
 		}
 	}
--- a/plugin/handlergen.go
+++ b/plugin/handlergen.go
@@ -614,7 +614,7 @@ func (p *OrmPlugin) generateStrictUpdateHandler(message *generator.Descriptor) {
 	if getMessageOptions(message).GetMultiAccount() {
 		p.generateAccountIdWhereClause()
 	}
-	p.P(`count := 1`)
+	p.P(`var count int64`)
 	// add default ordering by primary key
 	ormable := p.getOrmable(typeName)
 	if p.hasPrimaryKey(ormable) {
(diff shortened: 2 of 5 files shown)


example/feature_demo/demo_service.pb.gorm.go
```diff
@@ -354,13 +354,16 @@ func DefaultStrictUpdateIntPoint(ctx context.Context, in *IntPoint, db *gorm1.DB
...
 			return nil, err
 		}
 	}
-	if hook, ok := interface{}(&ormObj).(IntPointORMWithBeforeStrictUpdateSave); ok {
-		if db, err = hook.BeforeStrictUpdateSave(ctx, db); err != nil {
+	if err = db.Set("gorm:association_autoupdate", false).Set("gorm:association_autocreate", false).Set("gorm:association_save_reference", false).Save(&ormObj).Error; err != nil {
+		return nil, err
+	}
+	if hook, ok := interface{}(&ormObj).(IntPointORMWithBeforeStrictUpdateCleanup); ok {
+		if db, err = hook.BeforeStrictUpdateCleanup(ctx, db); err != nil {
```
Evgeniy-L: I see that the order of hooks is changed. Looks like a breaking change. Is it intentional?

dkukharau: The order of operations is changed, so the order of hooks is changed accordingly. I don't see obvious breaking here, though in some cases it can break things.

johnsdouglass:
Minor thing I noticed: The generateAccoundIdWhereClause method name in protoc-gen-gorm/plugin/handlergen.go has a typo. It should be called generateAccountIdWhereClause. Maybe not worth fixing.

plugin/handlergen.go
```diff
@@ -619,12 +630,13 @@ func (p *OrmPlugin) generateStrictUpdateHandler(message *generator.Descriptor) {
 		p.P(`return nil, err`)
 		p.P(`}`)
 	}
+	p.generateBeforeHookCall(ormable, "StrictUpdateSave")
+	p.P(`if err = db.Set("gorm:association_autoupdate", false).Set("gorm:association_autocreate", false).Set("gorm:association_save_reference", false).Save(&ormObj).Error; err != nil {`)
```
Calebjh: Is this mostly to lock the table? Is this the clearest way to do that? It seems redundant (the update of the base object) with the db.Save below.

plugin/handlergen.go
```diff
@@ -28,6 +28,14 @@ func (p *OrmPlugin) generateDefaultHandlers(file *generator.FileDescriptor) {
 	}
 }
 
+func (p *OrmPlugin) generateAccoundIdWhereClause() {
```
Calebjh: Per John D. should be AccountId instead of AccoundId, also appears where called in the code.

plugin/handlergen.go
```diff
@@ -470,13 +477,15 @@ func (p *OrmPlugin) generateListHandler(message *generator.Descriptor) {
...
+	if getMessageOptions(message).GetMultiAccount() {
+		p.generateAccoundIdWhereClause()
+	}
 	p.generateBeforeListHookCall(ormable, "ApplyQuery")
 	p.P(`db, err = `, p.Import(tkgormImport), `.ApplyCollectionOperators(ctx, db, &`, ormable.Name, `{}, &`, typeName, `{}, `, f, `,`, s, `,`, pg, `,`, fs, `)`)
 	p.P(`if err != nil {`)
 	p.P(`return nil, err`)
 	p.P(`}`)
 	p.generateBeforeListHookCall(ormable, "Find")
-	p.P(`db = db.Where(&ormObj)`)
```
Calebjh: Is there a compelling reason to remove this? I know that using the &ormObj from ToOrm() can cause issues when ToOrm() initializes a child object, but skipping it like this seems like it could also have consequences.

dkukharau: Initially this line was added here in order to apply AccountId condition to the List query clause. I decided to change it to a more clear way with `p.generateAccoundIdWhereClause()`.
But now I'm recalling that that was used by apps for applying some additional(non AccountId) filters(e.g. zone type). Is that true? If so, then I'll revert it.

plugin/handlergen.go
dkukharau: The thing is that we want to save all associations after updating the base object and deleting all associated child objects. I don't see a concise, clear way of doing this with GORM for now. We could save associations one-by-one, and that was something I was trying to achieve in
 my previous commit, but the difficulty is that we would need to save all associations(along with non-proto(included) ones), and this is a non-trivial, ambiguous thing to implement from the plugin perspective. So I ended up with the current solution.

plugin/handlergen.go
dkukharau: Will fix this, thanks.

plugin/handlergen.go
Calebjh: Is there an option to just 'lock' the table or entry somehow, or would that be no more clear or efficient than this update?

plugin/handlergen.go
Calebjh: This is a two sided issue. Using `db.Where(&ormObj)` allows for filtering on fields filled set in the ToOrm hooks, but its use here (and possibly elsewhere) also restricts ToOrm in an unexpected way. Setting a synthetic object in ToOrm() can cause these queries to fail (I don't have the exact error handy), because you can't filter on an object that has associations.

I'd say the new approach is the better one, but I think it will also cause some issues when upgrading to this code (potentially hard to detect issues). 
If we want to do this more 'breaking' change, can we cut a release of gentool with the fix in atlas-gen-validate before including this in the next version? @Evgeniy-L

plugin/handlergen.go
dkukharau: After thinking a bit on this I realized that we can lock a row while performing count operation. SELECT FOR UPDATE is used to do this.

PR #134, opened 2018-10-16
Add function and service generation for a multiple delete

Calebjh:
I asked @dmacthedestroyer what he thought of this idea a while back, but didn't get around to trying to include it until now.

The idea is to delete multiple entries simultaneously by passing a list of IDs instead of a single ID, then using `WHERE id in (?)` instead of `WHERE id = ?`.

With the grpc-gateway this requires a body in the delete request which requires this flag`--grpc-gateway_out=allow_delete_body=true,...:...` in the protoc invocation.

Comments, thoughts, concerns or reviews?


plugin/servergen.go
```diff
@@ -363,6 +369,60 @@ func (p *OrmPlugin) followsDeleteConventions(inType generator.Object, outType ge
...
+		p.P(`return out, nil`)
+		p.P(`}`)
+		p.generatePreserviceHook(service.ccName, method.baseType, method.ccName)
+		p.generatePostserviceHook(service.ccName, method.baseType, p.TypeName(method.outType), method.ccName)
+	} else {
+		p.generateEmptyBody(method.outType)
+	}
+}
+
+func (p *OrmPlugin) followsDeleteSetConventions(inType generator.Object, outType generator.Object, method *descriptor.MethodDescriptorProto) (bool, string) {
```
dmacthedestroyer: so the convention is the method must have an `ids` field, and be associated with an ormable type with a primary key?

This is likely _mostly_ good enough, but you could add further constraints on the `ids` field -- it must be repeated and be a primitive type, for instance

Calebjh:
This will be a new rpc name convention, so @dkukharau, do you have an opinion on the best name for it? @dmacthedestroyer proposes `DeleteSet` (initial choice) and `DeleteMany` are both good options.

Calebjh:
There's a use case where we have a self-referencing table, and trying to delete multiple objects "simultaneously" becomes sequentially and has errors occurring if the order they're deleted in doesn't match the constraints.

plugin/handlergen.go
```diff
@@ -465,42 +465,41 @@ func (p *OrmPlugin) generateDeleteSetHandler(message *generator.Descriptor) {
...
+	p.generateBeforeDeleteSetHookCall(ormable)
 	if getMessageOptions(message).GetMultiAccount() {
 		p.P(`acctId, err := `, p.Import(authImport), `.GetAccountID(ctx, nil)`)
 		p.P(`if err != nil {`)
 		p.P(`return err`)
 		p.P(`}`)
 		p.P(`err = db.Where("account_id = ? AND `, pkName, ` in (?)", acctId, keys).Delete(&`, ormable.Name, `{}).Error`)
 	} else {
-		p.P(`err := db.Where("`, pkName, ` in (?)", keys).Delete(&`, ormable.Name, `{}).Error`)
+		p.P(`err = db.Where("`, jgorm.ToDBName(pkName), ` in (?)", keys).Delete(&`, ormable.Name, `{}).Error`)
```
dkukharau: `column` tag should be considered like in `DefaultList`.

Evgeniy-L:
@Calebjh, I see that there are no updates for a long time in this PR. What is the next step for it?

PR #139, opened 2018-11-20
Fix Golang IDE syntax complaint 

burov:
Fix syntax error in template


Evgeniy-L:
It is a bit weird to see so many changes in *.pb.go files. Looks like different versions of protoc are used. We should be consistent in terms of protoc version at least across the team.

Evgeniy-L:
Per discussion agreed that we should have *.pb.go generated by gentool

PR #141, opened 2018-11-26
Set CREATED status in autogenerated Create methods

dkukharau:


Calebjh:
This is a good change, I've been considering it for a while (and thinking about how to incorporate similar status setting for things like "NOT FOUND").
I think it should be after the post-hook though, in case that returns an error that cancels or rolls back the transaction, and I think we may want a "gateway=true" option during generation or something like that to trigger the generation of this and the `gateway.SetCreated` that happens during the `Update` operation.

dkukharau:
- Things like "NOT FOUND" are handled by [error mappers](https://github.com/infobloxopen/atlas-app-toolkit/tree/master/errors) functionality. This PR is for cases when no error is raised.
- I've put this code before the hook to make it possible to override the status in custom code. If there is error in hook, the status set by `gateway.SetCreated` will be ignored and set according to error type.
- I'm ok with `gateway` option.

plugin/handlergen.go
```diff
@@ -706,9 +706,11 @@ func (p *OrmPlugin) generateStrictUpdateHandler(message *generator.Descriptor) {
 	p.P(`return nil, err`)
 	p.P(`}`)
 
-	p.P(`if count == 0 {`)
-	p.P(`err = `, p.Import(gatewayImport), `.SetCreated(ctx, "")`)
-	p.P(`}`)
+	if p.gateway {
+		p.P(`if count == 0 {`)
```
Calebjh: I believe this is the only use of `count`, so the part where it's declared and set above can also be wrapped in a similar `if`. I don't consider that critical though.

PR #150, opened 2018-12-28
Support for PQ arrays in Postgres

chrusty:
This allows protoc-gen-gorm to support the array types available in github.com/lib/pq (but only when engine=postgres):
- BoolArray
- Float64Array
- Int64Array
- StringArray

An example proto (plus generated code and `protoc` command) is also provided.


example/postgres_arrays/postgres_arrays.pb.gorm.go
```diff
@@ -0,0 +1,469 @@
...
+	var err error
+	if prehook, ok := interface{}(m).(ExampleWithBeforeToORM); ok {
+		if err = prehook.BeforeToORM(ctx, &to); err != nil {
+			return to, err
+		}
+	}
+	to.Id = m.Id
+	to.Description = m.Description
+	for _, v := range m.ArrayOfBools {
+		to.ArrayOfBools = append(to.ArrayOfBools, v)
```
Calebjh: (for all types of array)
Since we know in advance the size of the slice, I think it would be more efficient to `make` the slice of the length `n` as opposed to just appending `n` times which may result in multiple re-allocations. I'm not positive how the typing would work, but that might also allow for using the builtin `copy`?

chrusty: Hi @Calebjh, I've made the changes you requested. The typing works both ways, and the generated code has slightly less lines now too.

README.md
```diff
@@ -137,6 +137,13 @@ Within the proto files, the following types are supported:
   invocation) or between packages. All associations can be generated properly
   within the same package, but cross package only the belongs-to and many-to-many
   will work.
+- some repeated types can be automatically handled for Postgres by github.com/lib/pq, and
+  as long as the engine is set to postgres then to/from mappings will be created (see the
+  example called `postgres_arrays):
```
Calebjh: You've got a dangling \` here. Also, I think it would be better to have a link, like on line 103 & 105 (although it appears those links are broken right now and should probably be `example/feature_demo/*.proto`).

plugin/plugin.go
```diff
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	typeMessage = 11
-	typeEnum    = 14
+	allowStringArrays = true
```
Calebjh: This variable doesn't seem to be used anywhere else

README.md
chrusty: Alright, I changed this to provide the full path as a link to the postgres-arrays example (as this is the one using these arrays).

PR #158, opened 2019-02-12
Resurrect StrictUpdate concurrency fix

dkukharau:
Some time ago `SELECT FOR UPDATE` statement was added to prevent concurrency issues in StrictUpdate handler. At some point it was hidden by `gateway` option, but it shouldn't depend on it.


Evgeniy-L:
Looks good to me.
@Calebjh, can you please review this PR?

PR #160, opened 2019-02-14
Postgres DB Time type support

YReshetko:
This is the a limited implementation of time support for Postgress DB. 
Based on assumption that we can simply save and retrieve time as a string. Validation is done when we unmarshal data and during ORM mapping.

Supported time format:
`yyyy-MM-ddThh:mm:ssZ`
`hh:mm:ss`


types/time_only_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func TestParseValue(t *testing.T) {
+	cases := []struct {
+		value         uint64
+		str           string
+		expectedError error
+	}{
+		{0, "0000-00-00T00:00:00Z", nil},
+		{86399, "0000-00-00T23:59:59Z", nil},
+		{7158, "0000-00-00T01:59:18Z", nil},
```
Calebjh: Maybe some negative test cases, too?

YReshetko: Thanks for comment.

Updated test cases and added a few errors during time parsing

plugin/plugin.go
```diff
@@ -319,6 +320,8 @@ func (p *OrmPlugin) parseBasicFields(msg *generator.Descriptor) {
 				} else {
 					fieldOpts.Tag = tagWithType(tag, "varchar(48)")
 				}
+			} else if rawType == protoTimeOnly {
+				fieldType = "string"
```
dkukharau: Could you also include `type:time` tag to the generated struct field(like we do for other types): it may be useful for those who create their schema with `gorm.AutoMigrate`.
```golang
fieldOpts.Tag = tagWithType(tag, "time")
```

YReshetko:
> Greetings to Gomel! Thanks a lot for your contribution!
> I'm sorry for the long delay: I was hoping to find a bit more general solution for the issue, but your approach looks fine to me now, although I left some comments.
> P.S. Could you also update `example/feature_demo/demo_types.proto`(probably `TypeWithID` message) to demonstrate the usage of `TimeOnly`? It would also show that generated code is correct(or at least compiles).

Hi Daniil,

Thanks for detailed review, the PR is updated according your comments.
Apart from `TypeWithID` I've updated `TestTypes` as well to show marshaling process.

Could you also check if the `fieldOpts.Tag` added correctly. Tag is added to [ORM model](https://github.com/infobloxopen/protoc-gen-gorm/blob/1699612d6d66f3a45486f45ef12da2fa5cd7e012/example/feature_demo/demo_types.pb.gorm.go#L178) Is this what you expect to have?

Best Regards

PR #162, opened 2019-02-22
Handle DeletedAt

dkukharau:
`DeletedAt` has a special meaning in GORM. In order to work correctly it should be of `*time.Time` type.


plugin/plugin.go
```diff
@@ -772,14 +775,30 @@ func (p *OrmPlugin) generateFieldConversion(message *generator.Descriptor, field
...
 				p.P(`}`)
 			} else {
-				p.P(`if to.`, fieldName, `, err = `, p.Import(ptypesImport), `.TimestampProto(m.`, fieldName, `); err != nil {`)
+				var pointer string
+				if fieldName == "DeletedAt" {
+					pointer = "*"
+				}
+				if pointer != "" {
+					p.P(`if m.`, fieldName, ` != nil {`)
+				}
```
Calebjh: It seems to me that this block can be merged with the one above.

PR #163, opened 2019-02-25
Improve error raising

dkukharau:
Improve client's visibility on what exactly error is rased in pgg code.


plugin/imports.go
```diff
@@ -48,6 +48,7 @@ var (
 	queryImport    = "github.com/infobloxopen/atlas-app-toolkit/query"
 	gatewayImport  = "github.com/infobloxopen/atlas-app-toolkit/gateway"
 	pqImport       = "github.com/lib/pq"
+	gerrorsImport  = "github.com/infobloxopen/protoc-gen-gorm/plugin/errors"
```
Calebjh: I think I'd prefer for this to live in `.../protoc-gen-gorm/errors` instead of `.../protoc-gen-gorm/plugin/errors`

PR #166, opened 2019-04-26
Support NULL timestamps.

kd7lxl:
I'm having vendor trouble, but this does seem to work otherwise.


dkukharau:
@Calebjh we should be careful with this proposal since it's a breaking change. Probably an alternative exists, e.g. to add a field-level option which would define if pointer should be generated.

Evgeniy-L:
Changes for `user.pb.gorm.go` are missing in this PR

kd7lxl:
Yes, it is a breaking change, so under SemVer it will require a major version increment. `google.protobuf.Timestamp` is designed to be a nil-able field, so this is the correct change. It should have been `*time.Time` all along. The readme warns that breaking changes are expected, so this should be acceptable.

Does `protoc-gen-gorm` have an existing annotation for requesting that a field allow `NULL` or `NOT NULL`? That could be used to explicitly request `*time.Time`.

Calebjh:
Also, according to @kd7lxl, GORM does handle the `UpdatedAt` and `CreatedAt` fields properly as pointers. I think this change makes the proto Timestamp more versatile, since you can still set `*time.Time` to Zero, but before had no way to set `time.Time` to be Nil.

kd7lxl:
> 2\. gorm has special handling for `CreatedAt`, `UpdatedAt`, `DeletedAt` fields.  I may expect that changing it to pointer may cause some issues.

This PR only modifies downstream generated code, not upstream handling of `CreatedAt`, `UpdatedAt`, and `DeletedAt` fields. How gorm handles those is untouched.

Evgeniy-L:
@kd7lxl, gorm has some special build-in handling for `CreatedAt`, `UpdatedAt`, and `DeletedAt` fields. For example `db.Save(&user)` with implicitly set updated_at to current time.

If you take a look at [Gorm Base Model definition](http://doc.gorm.io/models.html#conventions) you can see that they define it as:
```
type Model struct {
  ID        uint `gorm:"primary_key"`
  CreatedAt time.Time
  UpdatedAt time.Time
  DeletedAt *time.Time
}
```
I want to make sure that having `UpdatedAt *time.Time` in our code will not break this implicit gorm's logic as some of our services rely on this. This is why I am curious if this use case was tested.

kd7lxl:
> @kd7lxl, gorm has some special build-in handling for CreatedAt, UpdatedAt, and DeletedAt fields. For example db.Save(&user) with implicitly set updated_at to current time.

Looks like the generic field setter they use for all of this functionality can resolve pointers through reflection, so we shouldn't have to worry about breaking it with a pointer type.

https://github.com/jinzhu/gorm/blob/a6382da48500a7adfe8a3f75eedc89a34644f54f/callback_update.go#L50-L55

https://github.com/jinzhu/gorm/blob/a6382da48500a7adfe8a3f75eedc89a34644f54f/field.go#L38-L43

> If you take a look at [Gorm Base Model definition](http://doc.gorm.io/models.html#conventions)

Does `protoc-gen-gorm` support inheriting `Model`? How do I do it?

Evgeniy-L:
> Does `protoc-gen-gorm` support inheriting `Model`? How do I do it?

No, we do not support it. The link is only as an example

example/user/user.override.go
```diff
@@ -9,6 +9,6 @@ import (
 // us to customize conversion behavior. In this example, we set the User's Age
 // based on the Birthday, instead of storing it separately in the DB
 func (m *UserORM) AfterToPB(ctx context.Context, user *User) error {
-	user.Age = uint32(time.Now().Sub(m.Birthday).Hours() / 24 / 365)
+	user.Age = uint32(time.Now().Sub(*m.Birthday).Hours() / 24 / 365)
```
Evgeniy-L: This will cause compilation error! Please regenerate and add `user.pb.gorm.go` to this PR.

kd7lxl: Updated generated code.

PR #175, opened 2019-10-29
fix embedded structs

Adphi:


Adphi:
Nothing have moved here after some years... I am going to close this one.

PR #178, opened 2019-11-28
Feature/remove remove child associations

### LINKED ISSUE #177: association_autoupdate does not work with DefaultStrictUpdate
The generated default server uses `DefaultStrictUpdate`. This deletes all children of the item you are trying to patch. 
e.g
```
filterProjects := ProjectORM{}
if ormObj.Id == 0 {
    return nil, errors1.EmptyIdError
}
filterProjects.ClientId = new(uint32)
*filterProjects.ClientId = ormObj.Id
if err = db.Where(filterProjects).Delete(ProjectORM{}).Error; err != nil {
    return nil, err
}
```

I do not understand the thinking behind this.

This means that https://github.com/infobloxopen/protoc-gen-gorm/blob/master/README.md#customization trying to control the behaviour of associations wont work as you have no control over whether children are deleted.

When using 
```
[(gorm.field).has_many = {
        association_autoupdate: false
}]
```
children should not be effected or in my opinion better yet, children should never be effected as this is a un-foreseen side effect and this logic should be left to the implementer to control 

Also please note this only "deletes children" when using deleted_at for soft delete.

mewis:
Removes `removeChildAssociations` as this is a unwanted side effect
fixes #177


mewis:
Please note the tests are failing dues to 
```
gorm/vendor/github.com/golang/protobuf/proto".ProtoPackageIsVersion3
FAIL	github.com/infobloxopen/protoc-gen-gorm/example/feature_demo [build failed]
```

I resolved this my updating the version of `github.com/golang/protobuf` to `v1.3.2` and this fixed the tests locally but Im not sure we this still fails with travis. maybe there is  not a fresh go install in your pipeline?

zdeeb-cbs:
Is there an update on this?

Calebjh:
Hey @zdeeb-cbs, December was pretty busy for me (and January may not be much better), but I still want to add a toggle option for this, like @mewis suggested in #177 when I'm able to find the time. Sorry for the slow turnaround.

mewis:
Hi @Calebjh I am seeing more and more the reason why we would want this as a toggle rather than a blanket rule either way.
Probably  something like
```
message MyMessage {
    option (gorm.opts) = {
        ormable: true,
    };
    uint32 id = 1 [(gorm.field).tag = {primary_key: true}];
    Assoc assoc = 2 [(gorm.field).belongs_to = {
        association_autoupdate: false,
        associate_mode: "append/replace/delete/clear"
    }];
```
This will correspond with https://gorm.io/docs/associations.html#Append-Associations.

What do you think?

mewis:
I have taken the idea I put above here
https://github.com/infobloxopen/protoc-gen-gorm/pull/187
I'll close this once I get feedback

PR #181, opened 2020-02-21
Add span

rdv821:


mphillips-infoblox:
Are you able to annotate the span with the input parameter values at the start of the method call and then with the result values before it returns?

rdv821:
> Are you able to annotate the span with the input parameter values at the start of the method call and then with the result values before it returns?

Thanks for comment. 
Annotations was added, please review.

example/feature_demo/demo_multi_file_service.proto
```diff
@@ -21,6 +21,6 @@ service BlogPostService {
     rpc Read(ReadAccountRequest) returns (ReadBlogPostsResponse);
 
     option (gorm.server) = {
-        autogen: true
+        autogen: true,
```
Evgeniy-L: can we revert this?

mphillips-infoblox:
What do you think of changing the option to "with_tracing" instead of "with_span"?  More generic description of what we are actually doing, instead of the implementation detail.

plugin/servergen.go
```diff
@@ -136,6 +136,20 @@ func (p *OrmPlugin) generateDefaultServer(file *generator.FileDescriptor) {
...
+	withSpan := getServiceOptions(service.ServiceDescriptorProto).WithSpan
+	if withSpan != nil && *withSpan {
+		rawVarName := fmt.Sprint("rawParameter", strings.Title(key))
+		errVarName := fmt.Sprint("errMarshaling", strings.Title(key))
+		p.P(rawVarName, `, `, errVarName, ` := `, p.Import(encodingJsonImport), `.Marshal(`, value, `)`)
+		p.P(`if `, errVarName, ` != nil {`)
+		p.P(`span.Annotate([]`, p.Import(ocTraceImport), `.Attribute{`, p.Import(ocTraceImport), `.StringAttribute("`, key, `", "")}, "`, message, `")`)
+		p.P(`return nil, `, errVarName)
+		p.P(`}`)
+		p.P(`span.Annotate([]`, p.Import(ocTraceImport), `.Attribute{`, p.Import(ocTraceImport), `.StringAttribute("`, key, `", string(`, rawVarName, `))}, "`, message, `")`)
```
burov: Should we instead of generating this piece of code twice for each repository generate generalised function once and use it in all methods?

mphillips-infoblox: Refactoring should be fine, provided you add the correct error handling code for the span.

rdv821:
> Looking very good. One thing I noticed, is that the generated code does not seem to include span error handling for any of the custom interceptors for the DB calls. i.e. In the error handlers for the Before* and After* calls.

Error handlers for Before* and After* calls was added

plugin/servergen.go
```diff
@@ -111,6 +111,9 @@ func (p *OrmPlugin) generateDefaultServer(file *generator.FileDescriptor) {
 			p.P(`DB *`, p.Import(gormImport), `.DB`)
 		}
 		p.P(`}`)
+		p.generateSpanInstantiationMethod(service)
+		p.generateSpanErrorMethod(service)
+		p.generateSpanResultMethod(service)
```
burov: Should we instead of checking ```withSpan``` flag in every function verify it's once there

plugin/servergen.go
```diff
@@ -136,29 +139,84 @@ func (p *OrmPlugin) generateDefaultServer(file *generator.FileDescriptor) {
 	}
 }
 
+func (p *OrmPlugin) generateSpanInstantiationMethod(service autogenService) {
+	withSpan := getServiceOptions(service.ServiceDescriptorProto).WithTracing
+	if withSpan != nil && *withSpan {
+		p.P(`// spanCreate ...`)
+		p.UsingGoImports(stdFmtImport)
+		p.P(`func (m *`, service.GetName(), `DefaultServer) spanCreate(ctx context.Context, in interface{}, methodName string) (*`, p.Import(ocTraceImport), `.Span, error) {`)
```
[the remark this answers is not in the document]
rdv821: This method is not direct but refers to the service. context binding is important in this case so that the method does not remain in limbo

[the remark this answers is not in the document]
burov: I don't understand what means for remain in limbo?

PR #184, opened 2020-03-30
Feature/allow multiple crud same service

mewis:
This feature will allow the auto generation of multiple crud methods per service.

An example of this would be that you have a service for managing all thing a client can do

The client might be able to create clients, users, items etc

this way we can in the same service have
```
service MyService {

    rpc CreateClient ( CreateClientRequest ) returns ( CreateClientResponse ) {}
    rpc CreateProject ( CreateProjectRequest ) returns ( CreateProjectResponse ) {}
    rpc CreateItem( CreateItemRequest ) returns ( CreateItemResponse ) {}
}
```

p.s apologies for the commits, I some how managed to merge another feature into this branch, this has been undone


example/feature_demo/demo_service.pb.gorm.go
```diff
@@ -1000,10 +1000,37 @@ type IntPointServiceIntPointWithAfterList interface {
...
 func (m *IntPointServiceDefaultServer) ListSomething(ctx context.Context, in *google_protobuf2.Empty) (*ListSomethingResponse, error) {
-	out := &ListSomethingResponse{}
+	db := m.DB
+	if custom, ok := interface{}(in).(IntPointServiceSomethingWithBeforeListSomething); ok {
+		var err error
+		if db, err = custom.BeforeListSomething(ctx, db); err != nil {
+			return nil, err
+		}
+	}
+	res, err := DefaultListSomething(ctx, db)
```
Calebjh, line 1010: I don't think that this function exists, which means that the way this call is generated would have to be fixed.

mewis: I might be missing something but I think this function does exist.
I can see the rpc defined in  feature_demo/demo_service.proto
And the function is generated in /feature_demo/demo_service.pb.gorm.go (line 642)
It looks like this function existed before my changes just was never used

mewis: Hi @Calebjh any thoughts on the above?

mewis:
@Calebjh maybe you could give me a brief on that the issue was before so I can see if I can resolve it without loosing the ability to have multiple create etc rpc in the same service. Cheers

PR #187, opened 2020-05-01
Feature/association handler

mewis:
This will add the ability to specify what action to take with a child association.

The idea is that you can specify at the message level wether you should replace or append associations.

The default is clear which should mean this works as before for anyone else


mewis:
Yeah, I think I know why that is, I'll look to fixing it. I am having issues running tests though. Don't know if you are having issues with this.
"module google.golang.org/grpc@latest found (v1.29.1), but does not contain package google.golang.org/grpc/transport"

mewis:
I have amended this so its replace by default.

Calebjh:
I'm sorry, my last comment wasn't very clear. 
The reason it was creating the custom code that deleted from the child table was because the GORM association `Replace` option doesn't actually remove them. For some applications this could be the desired behavior (not removing the DB objects, just the FKs) so I want to support these GORM association options, but with this change, what was the default behavior is removed as an option.

mewis:
Sorry for late response.
I will look at this asap, I guess I was thinking it would be nice to be as close to what gorm does as possible but I understand this change would be a breaking change which is not great.

mewis:
I have gone a slightly different angle as I was worried about breaking backwards compatibility.
I have added the tag `gorm_association_defaults` when this is set then you can use gorm's default way of handling associations. If this is not set then this works as before.

This way this feature does not collide with how gorm works so will be clear when looking at gorm documentation

mewis:
@Calebjh this should be ready for a review now.
It works that if you set one of `append`, `replace`, `clear` then it works the same way gorm does, if not set then it works as before
For many to many default is `replace` as you did not handle these before. I can change this if required.

plugin/handlergen.go
```diff
@@ -781,10 +781,73 @@ func (p *OrmPlugin) isFieldOrmable(message *generator.Descriptor, fieldName stri
...
+
+	if field.GetHasMany() != nil || field.GetHasOne() != nil || field.GetManyToMany() != nil {
+		var assocHandler string
+		switch {
+		case field.GetHasMany() != nil:
+			switch {
+			case field.GetHasMany().GetAppend():
+				assocHandler = "Append"
+			case field.GetHasMany().GetClear():
+				assocHandler = "Clear"
```
Calebjh, line 807:
```suggestion
			case field.GetHasMany().GetClear():
				assocHandler = "Clear"
			case field.GetHasMany().GetAppend():
				assocHandler = "Append"
```
For consistency with the other switches, it should be in this order

plugin/plugin.go
```diff
@@ -572,6 +581,15 @@ func (p *OrmPlugin) renderGormTag(field *Field) string {
...
 		gormRes += fmt.Sprintf("preload:%s;", strconv.FormatBool(*preload))
 	}
+	if clear != nil {
+		gormRes += fmt.Sprintf("clear:%s;", strconv.FormatBool(*clear))
+	}
+	if replace != nil {
+		gormRes += fmt.Sprintf("replace:%s;", strconv.FormatBool(*replace))
+	}
+	if append != nil {
+		gormRes += fmt.Sprintf("append:%s;", strconv.FormatBool(*append))
```
Calebjh, line 591: I may be missing it, I don't see these as gorm tags here http://v1.gorm.io/docs/models.html#Struct-tags-for-Associations, and aren't these options mutually exclusive, so an `else if` chain might make more sense?

mewis: Hi @Calebjh, yeah they are not gorm tags but rather instructions for us on how to deal with associations.
I probably didn't need have them in `message GormTag `

mewis: I have removed from gormTag message and used else if chain.
Please let me know if you think this should be done somewhere else.

PR #215, opened 2021-09-01
remove gogo

khous:


go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/infobloxopen/protoc-gen-gorm
 
 require (
 	github.com/denisenkom/go-mssqldb v0.9.0 // indirect
-	github.com/gogo/protobuf v1.3.2
+	github.com/gogo/protobuf v1.2.1
```
drewwells, line 5: missed one

example/feature_demo/demo_multi_file.pb.go
```diff
@@ -1,12 +1,13 @@
...
-// 	protoc-gen-go v1.27.1
-// 	protoc        v3.15.7
+// 	protoc-gen-go v1.23.0
+// 	protoc        v3.17.1
 // source: feature_demo/demo_multi_file.proto
 
 package example
 
 import (
+	proto "github.com/golang/protobuf/proto"
```
binjip978, line 10: I don't think this is right, because it uses deprecated go bindings for protocol buffers, instead we should use  google.golang.org/protobuf and also I see version downgrade from protoc-gen-go v1.27.1 to v1.23.0

example/user/user.pb.gorm.go
```diff
@@ -3,6 +3,7 @@ package user
 import (
 	context "context"
 	fmt "fmt"
+	timestamp "github.com/golang/protobuf/ptypes/timestamp"
```
binjip978, line 6: the same thing, it should not use old proto bindings timestamppb "google.golang.org/protobuf/types/known/timestamppb" should be used instead

binjip978:
why you are deleting form examples files and tests for them? Example files are a good references to see changes that can be introduced by code generation.

example/feature_demo/github.com/infobloxopen/protoc-gen-gorm/example/feature_demo/demo_multi_file.pb.gorm.go
```diff
@@ -7,15 +7,11 @@ import (
 	gorm1 "github.com/infobloxopen/atlas-app-toolkit/gorm"
 	errors "github.com/infobloxopen/protoc-gen-gorm/errors"
 	gorm "github.com/jinzhu/gorm"
-	go_uuid "github.com/satori/go.uuid"
 	field_mask "google.golang.org/genproto/protobuf/field_mask"
 )
 
 type ExternalChildORM struct {
-	Id                  string
```
binjip978, line 15: Primary* stuff is missing because types proto was not used in code generation.

binjip978:
I still see some examples are missing, the current result in main can be achieved by 
```bash
make gentool
make generate-gentool
```
but if you trying to do it without hacked gentool, it will be tricky, i tried but have problems with types.proto example and revert to docker approach.

khous:
> I still see some examples are missing, the current result in main can be achieved by
> 
> ```shell
> make gentool
> make generate-gentool
> ```
> 
> but if you trying to do it without hacked gentool, it will be tricky, i tried but have problems with types.proto example and revert to docker approach.

Thanks, I'll try it that way

khous:
> I still see some examples are missing, the current result in main can be achieved by
> 
> ```shell
> make gentool
> make generate-gentool
> ```
> 
> but if you trying to do it without hacked gentool, it will be tricky, i tried but have problems with types.proto example and revert to docker approach.

I ran with generate gentool. I see the missing example files showing up now. I was running into trouble getting the tests to work then I realized that I delete demo_types_gorm.custom.go

example/user/user.override.go
```diff
@@ -1,14 +0,0 @@
-package user
```
Calebjh, line 1: Probably not intended to be deleting this file, either

khous: thanks, got it

.github/workflows/test-gorm.yml
```diff
@@ -0,0 +1,13 @@
...
+  #   branches: [ main ]
+  # pull_request:
+  #   branches: [ main ]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Unit Tests
+      run: make test
```
drewwells, line 13: 💯

Makefile
```diff
@@ -16,11 +16,19 @@ build: $(BUF)
 test: lint build
 	go test -v ./...
 
+delete-generated:
```
drewwells, line 19: Normally, this is called clean
```suggestion
clean:
```

buf.yaml
```diff
@@ -16,7 +16,7 @@ lint:
     - protoc-gen-openapiv2
     - github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options/openapiv2.proto
     - google
-    - github.com/gogo
+    # - github.com/gogo
```
drewwells, line 19:
```suggestion
```
I like deleted better

example/feature_demo/demo_types_test.go
```diff
@@ -1,111 +0,0 @@
-package example
```
drewwells, line 1: Did this test get moved somewhere else and github lost it?

khous: probably got deleted. I'll do another lookaround to see what's gone

Makefile
Calebjh: There's another target called `clean` in Makefile.buf so they override each other if they're both named clean

Calebjh (approved):
Still some wonky paths for generated test files, but that was the case before, too, I think

PR #216, opened 2021-09-03
Fix a condition for file generation and the buf build configs

Calebjh:
Was making a minor change in plugin.go to restrict file generation to only files passed into for generation (otherwise it would sometimes generate files for included imports which didn't work right...)
But I ended up also fixing buf building, and removing the gogo references, superseding #215 I think


khous (approved):
Looks good from what I can tell

PR #237, opened 2022-07-25
Upgrade gORM to V2

MichaelSkuratovichNS:
#### `Upgrade gORM to V2`

This PR introduces new version of `gORM` in accordance to [release notes](https://gorm.io/docs/v2_release_note.html):
- Renames tags to CamelCase correct values
- Removes `association_autocreate`, `association_autoupdate`, `preload` tags from _generated code_, however replaces them with required code generation:
  - `association_autocreate` & `association_autoupdate` are replaced with `Omit(<field_name>` calls
  - `preload` is replaced with `Preload(<field_name>)` calls
-  `association_save_reference` is removed as it is missing equivalent in `gORMv2`
- Updates handling of `Jsonb` fields
- Changes Imports to related ones

_NOTE_: Currently this PR is dependent on [toolkit:gorm_v2 branch, which is not master](https://github.com/infobloxopen/atlas-app-toolkit/tree/gorm_v2), merges should be performed with @Calebjh agreement/review


MichaelSkuratovichNS:
Please do not merge this PR until discussion is finished and explicit note mentions it. This PR might depend on new changes in [toolkit](https://github.com/infobloxopen/atlas-app-toolkit/pull/331)

README.md
```diff
@@ -135,7 +135,7 @@ Within the proto files, the following types are supported:
   level.
 - custom wrapper type `gorm.types.JSONValue`, which wraps a string in protobuf
   containing arbitrary JSON and converts to `postgres.Jsonb` GORM type
-  (https://github.com/jinzhu/gorm/blob/master/dialects/postgres/postgres.go#L59)
+  (https://gorm.io/gorm/blob/master/dialects/postgres/postgres.go#L59)
```
Calebjh, line 138: The `/dialects` directory doesn't exist in GormV2, I don't think

plugin/plugin.go
```diff
@@ -876,8 +879,8 @@ func (b *ORMBuilder) parseBasicFields(msg *protogen.Message, g *protogen.Generat
 				fieldType = "*" + generateImport("Time", stdTimeImport, g)
 			} else if rawType == protoTypeJSON {
 				if b.dbEngine == ENGINE_POSTGRES {
-					typePackage = gormpqImport
-					fieldType = "*" + generateImport("Jsonb", gormpqImport, g)
+					typePackage = gtypesImport
```
Calebjh, line 882: Can we leverage the new JSON datatype from upstream, or is the custom one better (maybe simpler)? https://github.com/go-gorm/datatypes

example/user/user.proto
```diff
@@ -23,12 +23,12 @@ message User {
     uint32 age = 5 [(gorm.field).drop = true]; // synthetic field
     uint32 num = 6;
     CreditCard credit_card = 7; // has one
-    repeated Email emails = 8; // has many
-    repeated Task tasks = 9 [(gorm.field).has_many = {position_field: "priority" foreignkey_tag: {not_null: true}}];
+    repeated Email emails = 8 [(gorm.field).has_many = {disable_association_autocreate: true disable_association_autoupdate: true preload: true}]; // has many
```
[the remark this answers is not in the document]
MichaelSkuratovichNS: I believe it always was like this and had to be enabled like following: [gORMv1](http://v1.gorm.io/docs/preload.html#Auto-Preloading)

plugin/plugin.go
MichaelSkuratovichNS: I'm generally concerned over the fact, that we have a pointer to the struct. Thus I can assume, that some application code might rely on checking for `nil` field value, while new `JSON` datatype will hold `"null"` in case scanned from DB and we can't easily distinguish empty slice value from `nil` slice.

Calebjh (approved):
Seems okay, but I haven't tested any conversions to GormV2 personally

dpup:
Is there an intent to land this change any time soon? It would be great to get gorm2 support.

Thrun12:
Any update?

PR #255, opened 2023-12-05
feat: add support for clickhouse as the engine

alipourhabibi:
Add support for clickhouse as the engine for arrays and custom types specified in proto/types/types.proto file

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -911,16 +911,16 @@ func (b *ORMBuilder) parseBasicFields(msg *protogen.Message, g *protogen.Generat
 		} else if b.dbEngine == ENGINE_CLICKHOUSE && b.IsAbleToMakePQArray(fieldType) && field.Desc.IsList() {
 			switch fieldType {
 			case "bool":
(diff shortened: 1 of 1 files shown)


plugin/plugin.go
```diff
@@ -151,6 +152,8 @@ func New(opts protogen.Options, request *pluginpb.CodeGeneratorRequest) (*ORMBui
 
 	if strings.EqualFold(params["engine"], "postgres") {
 		builder.dbEngine = ENGINE_POSTGRES
+	} else if strings.EqualFold(params["engine"], "clickhouse") {
```
chinmayb, line 155: Since now we're introducing a new engine, its high time we modularize it by refactoring it. May be we can define an interface & include the methods which are different for different engines.

alipourhabibi: @chinmayb Should i start working on this new refactoring you talked about?

chinmayb: yes please

alipourhabibi: sure, i will open a new PR when i'm done with that

PR #263, opened 2024-02-29
feature-compartment

sbhagate-infoblox:


kumaya (commented):
LGTM. Run `go mod vendor` locally and commit `go.sum` file. Given changes in `go.mod` file; I expect changes in `go.sum` as well.

kumaya (approved):

PR #283, opened 2025-01-23
DDIDNS-1198 prevent dependency hell by splitting go mods

kanaksinghal:


chinmayb:
You may also need to do on top of 
https://github.com/infobloxopen/protoc-gen-gorm/releases/tag/v0.20.2

types/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/infobloxopen/protoc-gen-gorm/types
+
+go 1.23.4
+
+require (
+	github.com/golang/protobuf v1.5.4
+	github.com/google/go-cmp v0.6.0
+	github.com/lib/pq v1.10.9
```
kd7lxl, line 8: Looks like this might not actually be needed: https://github.com/infobloxopen/protoc-gen-gorm/blob/7402b542b5ddec9b0d47bdb28ae1d46642c0ef95/types/inet_test.go#L8
I don't see anything in this file that would depend on pq being registered.

Makefile
```diff
@@ -103,3 +103,10 @@ build-postgres-local:
 	-I./proto/ \
 	-I./third_party/proto/ \
 	example/postgres_arrays/postgres_arrays.proto --gorm_out="engine=postgres,enums=string,gateway:./example/postgres_arrays" --go_out=./example/postgres_arrays
+
+.PHONY: mod
+mod:
+	cd types && go mod tidy
+	cd example && go mod tidy
+	go mod tidy
```
kd7lxl, line 111: You can future proof this with
```suggestion
	find . -name go.mod -execdir go mod tidy \;
```

kd7lxl (approved):

types/go.mod
kanaksinghal: you are right, I removed and re-ran tests seems fine.

kanaksinghal:
> You may also need to do on top of https://github.com/infobloxopen/protoc-gen-gorm/releases/tag/v0.20.2

Thanks @chinmayb for bringing this up. I'll raise a Patch PR after this one merges.

chinmayb:
@kanaksinghal we can add a github action to check if go mod runs fine during PR

chinmayb (approved):

PR #286, opened 2025-02-04
DDIDNS-1198 prevent dependency hell (patch v0.21)

kanaksinghal:
by splitting go mods


go.work
```diff
@@ -0,0 +1,7 @@
+go 1.23.4
```
kanaksinghal, line 1:
```suggestion
go 1.16
```

chinmayb (approved):

go.work
kanaksinghal: removed go.work. since workspace is introduced in go v1.18.

chinmayb (approved):

